PR #6, opened 2025-05-09
[apps] refactor activity/event routing

aacebo:


Libraries/Microsoft.Teams.Apps.Testing/Events/App.cs
```diff
@@ -0,0 +1,28 @@
...
+public class TestMessageEvent : Event
+{
+    public required string Message { get; set; }
+}
+
+public static partial class AppExtensions
+{
+    public static App OnTestMessage(this App app, Action<TestPlugin, TestMessageEvent> handler)
+    {
+        return app.OnEvent("test.message", (plugin, @event) =>
```
heyitsaamir, line 15: is this "test." enforced? Or convention

Libraries/Microsoft.Teams.Apps/Activities/CommandActivity.cs
```diff
@@ -1,23 +1,19 @@
...
-public partial interface IRoutingModule
+public static partial class AppActivityExtensions
 {
-    public IRoutingModule OnCommand(Func<IContext<CommandActivity>, Task<object?>> handler);
-}
-
-public partial class RoutingModule : IRoutingModule
-{
-    public IRoutingModule OnCommand(Func<IContext<CommandActivity>, Task<object?>> handler)
+    public static App OnCommand(this App app, Func<IContext<CommandActivity>, Task<object?>> handler)
```
heyitsaamir, line 14: nice this is clean. Extending app like this

Libraries/Microsoft.Teams.Apps/Activities/Events/EventActivity.cs
```diff
@@ -1,23 +1,19 @@
...
-    public IRoutingModule OnEvent(Func<IContext<EventActivity>, Task<object?>> handler);
-}
-
-public partial class RoutingModule : IRoutingModule
-{
-    public IRoutingModule OnEvent(Func<IContext<EventActivity>, Task<object?>> handler)
+    public static App OnEvent(this App app, Func<IContext<EventActivity>, Task<object?>> handler)
     {
-        Router.Register(new Route()
+        app.Router.Register(new Route()
```
heyitsaamir, line 16: Help me understand this - so in typescript, events are emitted, but activities are routed. The main reason for this (i think) are that for events, we have more of a fire-and-forget system, whereas activities use a router and they may have multiple handlers that could get routed in order, and we might care about the result of the handlers to send back to the backend. 

In dotnet, it looks like everything is using the router

Libraries/Microsoft.Teams.Apps/Events/EventEmitter.cs
```diff
@@ -8,63 +8,75 @@ internal class EventEmitter
...
+    public async Task<object?> Emit(IPlugin plugin, string name, Event? @event = null, CancellationToken cancellationToken = default)
     {
-        var topic = Topics.TryGetValue(name, out Topic? value) ? value : [];
-        return topic.Emit(plugin, @event, cancellationToken);
+        if (!Topics.ContainsKey(name))
+        {
+            Topics[name] = [];
+        }
+
+        return await Topics[name].Emit(plugin, @event, cancellationToken);
```
heyitsaamir, line 80: Really dumb question, but if Topics[name] is an empty array, how can we call emit on it? Do we have an extension on this empty array?

heyitsaamir: Oh nvm i understand. It extends List, and has a function called emit got it that saves the handler

Libraries/Microsoft.Teams.Apps/AppPlugins.cs
```diff
@@ -26,36 +27,34 @@ public partial class App
...
+        plugin.Events += async (plugin, name, @event, token) =>
+        {
+            var eventType = new EventType(name);
+
+            await Events.Emit(plugin, $"{attr.Name}.{name}", @event, token);
+
+            if (eventType.IsBuiltIn && !eventType.IsStart)
+            {
+                return await Events.Emit(plugin, name, @event, token);
+            }
```
heyitsaamir, line 42: So if i have an event "Activity" in my plugin, how does the handler know what the payload will be for "Activity"?

heyitsaamir (approved):
Nothing blocking, just questions.
I like how everything is very similar to how stuff works in ts, conceptually. I understood most things.

Libraries/Microsoft.Teams.Apps.Testing/Events/App.cs
aacebo: its namespaced, so any events the test plugin emits will be prefixed with `test.`.

Libraries/Microsoft.Teams.Apps/Activities/Events/EventActivity.cs
aacebo: its the same in .net, the events are just emitted, whats confusing you is that this is the event actitivty which is an activity with type `event`.

Libraries/Microsoft.Teams.Apps/AppPlugins.cs
aacebo: each route has a way of casting the base activity context to its more specific activity type 
https://github.com/microsoft/teams.net/blob/main/Libraries/Microsoft.Teams.Apps/Routing/Route.cs#L56
https://github.com/microsoft/teams.net/blob/main/Libraries/Microsoft.Teams.Apps/Routing/Activities/CommandActivity.cs#L8

PR #17, opened 2025-05-16
Test coverage for Common package 

fallothakor:
HttpClient & HttpRequest  tests


Libraries/Microsoft.Teams.Common/Http/HttpClient.cs
```diff
@@ -73,37 +73,35 @@ protected HttpRequestMessage CreateRequest(IHttpRequest request)
...
-                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull
-            });
+                httpRequest.Content = JsonContent.Create(request.Body, options: new JsonSerializerOptions()
+                {
+                    DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull
+                });
+            }
         }
 
+        foreach (var kv in request.Headers)
```
fallothakor, line 95: Alex, I need to put code for setting Content header after setting up Httprequest.Content from body, without this "Content-" header were not set.

singhk97: Makes sense because `Content` would be overridden in the previous if statement

singhk97 (approved):
LGTM!

PR #74, opened 2025-06-11
Fthakor/sample apps

fallothakor:
Sampe App for ME functionality + fix
Added tests on FileConsentActivity + fix


Libraries/Microsoft.Teams.Apps/Activities/Invokes/MessageExtensions/SubmitActionActivity.cs
```diff
@@ -15,7 +15,7 @@ public class SubmitActionAttribute() : InvokeAttribute(Api.Activities.Invokes.Na
 
 public static partial class AppInvokeActivityExtensions
 {
-    public static App OnSubmitAction(this App app, Func<IContext<MessageExtensions.SubmitActionActivity>, Task<object?>> handler)
+    public static App OnMessageExtensionSubmitAction(this App app, Func<IContext<MessageExtensions.SubmitActionActivity>, Task<object?>> handler)
```
fallothakor, line 18: Hi Alex, I needed to make this change since there is another OnSubmitAction extention method on app, pls let me know if you think there might be other way to handle this.

Tests/Microsoft.Teams.Apps.Tests/Activities/Invokes/FileConsentActivityTests.cs
```diff
@@ -0,0 +1,127 @@
...
+    private readonly IToken _token = Globals.Token;
+    private readonly FileConsentController _controller = new();
+
+    public FileConsentActivityTests()
+    {
+        _app.AddPlugin(new TestPlugin());
+        _app.AddController(_controller);
+    }
+
+    //[Fact]
```
fallothakor, line 23: Alex, these tests kept failing even though onFileConsent is valid extension method. do you have quick pointers on what might be wrong here? Otherwise, would remove this test for now and will pick up after current task

Libraries/Microsoft.Teams.Apps/Activities/Invokes/MessageExtensions/SubmitActionActivity.cs
aacebo: thats fine, but if we are going to rename one we need to rename them all to stay consistent

Tests/Microsoft.Teams.Apps.Tests/Activities/Invokes/FileConsentActivityTests.cs
aacebo: @fallothakor in your controller handler your not calling `context.Next()` so the next handler will never be called

Issue #81, opened 2025-06-23
Add support for passing custom data to context through `App.Process()`


singhk97:
The context object supports holding [extra] (https://github.com/microsoft/teams.net/blob/main/Libraries/Microsoft.Teams.Apps/Contexts/Context.cs#L65) data. Users should be able to define static context extras on App creation time and dynamic context extra on a per-request (or activity) level basis.

Feature:
* Add support for passing extra data to context in the [`ISenderPlugin.Do`](https://github.com/microsoft/teams.net/blob/6e60e49eec703748c9a1b1df434c1ecd67174356/Libraries/Microsoft.Teams.Apps/Plugins/SenderPlugin.cs?plain=1#L10C18-L10C31) and [`App.Process()`](https://github.com/microsoft/teams.net/blob/b4fb261ef78741a3ac023785e13ad411a0d6e23d/Libraries/Microsoft.Teams.Apps/App.cs?plain=1#L338C1-L338C130) method.

singhk97:
![Image](https://github.com/user-attachments/assets/b3821914-99f1-49eb-9e43-1a7ec8d804d7)

Data flow diagram

singhk97:
Implemented with #82

PR #82, opened 2025-06-23
Add support for passing custom data to context through `App.Process()`

singhk97:
https://github.com/microsoft/teams.net/issues/81

### DIFF
--- a/Libraries/Microsoft.Teams.Apps/App.cs
+++ b/Libraries/Microsoft.Teams.Apps/App.cs
@@ -261,7 +261,7 @@ public partial class App
             Ref = reference,
             IsSignedIn = userToken is not null,
             OnNext = Next,
-            Extra = contextExtra,
+            Extra = contextExtra ?? new Dictionary<string, object>(),
             UserGraph = new Graph.GraphServiceClient(userGraphTokenProvider),
             CancellationToken = cancellationToken,
             ConnectionName = OAuth.DefaultConnectionName,
--- a/Tests/Microsoft.Teams.Apps.Tests/Activities/ActivityTests.cs
+++ b/Tests/Microsoft.Teams.Apps.Tests/Activities/ActivityTests.cs
@@ -58,7 +58,7 @@ public class ActivityTests
     }
 
     [Fact]
-    public async Task Should_Pass_ContextExtra_OnActivity_PluginConfigured()
+    public async Task Should_Pass_ContextExtra_OnActivity()
     {
         IDictionary<string, object>? extra = null;
         _app.OnActivity(context =>
@@ -74,6 +74,31 @@ public class ActivityTests
         var res = await _app.Process<TestPlugin>(_token, new MessageActivity(), contextExtraFromParameter);
 
         Assert.Equal(extra!["paramContextKey"], "value");
+    }
+
+    [Fact]
+    public async Task Should_Pass_ContextExtra_AcrossActivityHandlers()
+    {
+        
+        _app.OnActivity(async context =>
+        {
+            // Set the context in the first handler
+            context.Extra["key1"] = "value1";
+            // Call the next handler in the pipeline
+            await context.Next();
+        });
+
+        IDictionary<string, object>? extra = null;
+        _app.OnActivity(context =>
+        {
+            // Retrieve the context data set in the previous handler
+            extra = context.Extra;
+            return Task.CompletedTask;
+        });
+
+        var res = await _app.Process<TestPlugin>(_token, new MessageActivity());
+        Assert.Equal(System.Net.HttpStatusCode.OK, res.Status);
+        Assert.Equal(extra!["key1"], "value1");
     }
 
     [TeamsController]


heyitsaamir (dismissed):
Tests?

Libraries/Microsoft.Teams.Apps/AppBuilder.cs
```diff
@@ -95,6 +95,12 @@ public AppBuilder AddOAuth(string defaultConnectionName)
         return this;
     }
 
+    public AppBuilder AddContextExtra(IDictionary<string, object> extra)
```
aacebo, line 98: this is nice for global scenario, but I think they also want to be able to pass in custom data on each request, meaning we would need to also support this on a per request basis in the aspnetcore plugin and the `App.Process` method

aacebo (changes requested):

Libraries/Microsoft.Teams.Apps/AppBuilder.cs
singhk97: good catch - pushed commit

Libraries/Microsoft.Teams.Apps/App.cs
```diff
@@ -250,6 +252,19 @@ public async Task<Response> Process(ISenderPlugin sender, IToken token, IActivit
...
 
+        // merge extra data into context extras
+        var mergedExtra = new Dictionary<string, object>(ContextExtra);
+        foreach (var plugin in Plugins)
+        {
+            if (plugin.ContextExtra is not null)
+            {
+                foreach (var kvp in plugin.ContextExtra)
+                {
+                    mergedExtra[kvp.Key] = kvp.Value;
```
lilyydu, line 263: if there was already an entry, this would override right? should we add that as a note somewhere?

lilyydu (commented):
should we also add docs for this?

Libraries/Microsoft.Teams.Apps/Plugins/Plugin.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.Teams.Apps.Plugins;
 /// </summary>
 public interface IPlugin
 {
+    public IDictionary<string, object>? ContextExtra { get; set; }
```
heyitsaamir, line 11: should this be a `buildContextExtra`? since it might be different for each execution?

Libraries/Microsoft.Teams.Apps/App.cs
singhk97: once this PR gets merged i'll write some docs for this feature (possibly in [this](https://microsoft.github.io/teams-ai/typescript/essentials/app-basics#plugins) section) - as for this i'll add a comment

Libraries/Microsoft.Teams.Apps/Plugins/Plugin.cs
lilyydu: hmm if its per request, maybe we can somehow denote that?

for the `extra` part`, it might be a little trippy where extra denotes "additional" but it can also override old entries

aacebo: left a comment but this should be on a request basis and plugins are initialized on startup, we don't want things to be global we want to add a way to set this extra key/value data on a per-request basis via the existing `Do` sender method https://github.com/microsoft/teams.net/blob/main/Libraries/Microsoft.Teams.Apps/Plugins/SenderPlugin.cs#L45 (which should be an overload since some may not want to set that extra data)

aacebo (changes requested):
a few comments but good direction

rajan-chari:
If there are multiple instances of an App w/ round-robin load balancing, what's the expectation from a dev perspective about lifetime of Context-Extra?  For example, if auth happens in the middle of servicing an activity?

singhk97:
> If there are multiple instances of an App w/ round-robin load balancing, what's the expectation from a dev perspective about lifetime of Context-Extra? For example, if auth happens in the middle of servicing an activity?

The ContextExtra is a per-request level object. That means that it should be created on the fly when the app recieves an incomming request. And be disposed once the app has handled the request. This includes if auth happens in the middle of servicing an activity.

singhk97:
Latest update based on @aacebo's feedback:

* Removed the `IPlugin.ContextExtra` property since it introduces state where it's not needed. The context extra is meant to be a per-request level object - and adding it as a property can cause uninteded side effects such as the object perisisting beyond the current turn.
* Modified the `ISenderPlugin.Do` to take in context extra as a paramter since that's the entry-point to the app's activity processing method.

aacebo (commented):
can you also add a test case for setting a new key/value from a handler and asserting that key/value being present in the next chained handler?

singhk97:
For future reference:

@aacebo:
> plugins were designed to be out of sight out of mind, and I specifically made the decision to put context creation on the app not the plugins for several reasons, one of which being that even tho one plugin may receive the activity, any plugin can handle that activity, so it just doesn't make sense for the plugin to build the context itself. On top of that plugins can already do this via OnActivity and all the other lifecycle hooks
 
> devs should be interacting with plugins as little as possible, they were designed to be import and forget, they run and do things in the background but they should not be mutated by the dev in 99% of cases 
 
> the Context is the property of the App class, plugins can mutate it in the same way devs can in their handlers, but they do not create the context. If you did want to add some context factory (still strange since it is fully possible today via a route handler like you would with routers like express), it would need to be added to the App class, not a plugin.
 
> tldr, adding this stuff to the plugins would be a breach of separation of concerns and would cross the virtual boundary of responsibilities that the App and Plugins have.

PR #107, opened 2025-08-26
[API] add memberships for new channel data

aacebo:
https://github.com/microsoft/teams.net/issues/100


Libraries/Microsoft.Teams.Api/ChannelData.cs
```diff
@@ -80,6 +81,34 @@ public class ChannelData
...
+    [JsonPropertyName("sharedWithTeams")]
+    [JsonPropertyOrder(12)]
+    public IList<Team>? SharedWithTeams { get; set; }
+
+    /// <summary>
+    /// List of temas that a channel was unshared from
+    /// </summary>
+    [JsonPropertyName("unsharedFromTeams")]
+    [JsonPropertyOrder(13)]
+    public IList<Team>? UnSharedFromTeams { get; set; }
```
arocc, line 103: Hi @aacebo , just a nit that on bot side we don't capitalize the "S". Not sure if that would cause an issue? We have `UnsharedFromTeams`

arocc (approved):

Libraries/Microsoft.Teams.Api/ChannelData.cs
aacebo: won't make a difference since it will serialize and deserialize to `unsharedFromTeams`

PR #123, opened 2025-09-15
Fix sample launch settings

rajan-chari:
Fix launch settings for Echo and Lights samples


singhk97 (approved):

singhk97 (changes requested):
MCP and Tabs sample also need updating

Issue #109, opened 2025-09-02
Lifetime check validation fail in Development when using AddOpenAI


yuqizhou77:
When using AddOpenAI extension in Microsoft.Teams.AI SDK, I ran into a lifetime mismatch issue. Specifically, OpenAIChatPrompt is depends on services like ILogger, which is scoped in https://github.com/microsoft/teams.net/blob/32093fb5f2bb3f60ef49678704e42742df35846f/Libraries/Microsoft.Teams.Extensions/Microsoft.Teams.Extensions.Hosting/Microsoft.Teams.AI.Models.OpenAI.Extensions/ServiceCollection.cs#L60C8-L74C12. 
Therefore adding Controller in app.useTeams() cannot pass scope validation which is on by default in Development environment.

Steps to reproduce in Samples.light:

1.  Set  "ASPNETCORE_ENVIRONMENT" to "Development" in `Properties/launchSettings.json`
2.  Add  "OpenAI" section in appsettings.Development.json.
```
"OpenAI": {
    "ApiKey": "some key",
    "Model": "some model"
  }
```
4. dotnet build
5. dotnet run 

Errors popup in  `dotnet run`
```
Building...
Unhandled exception. System.AggregateException: Some services are not able to be constructed (Error while validating the service descriptor 'ServiceType: Samples.Lights.Controller Lifetime: Singleton ImplementationType: Samples.Lights.Controller': Cannot consume scoped service 'Microsoft.Teams.AI.Models.OpenAI.OpenAIChatPrompt' from singleton 'Samples.Lights.Controller'.)
 ---> System.InvalidOperationException: Error while validating the service descriptor 'ServiceType: Samples.Lights.Controller Lifetime: Singleton ImplementationType: Samples.Lights.Controller': Cannot consume scoped service 'Microsoft.Teams.AI.Models.OpenAI.OpenAIChatPrompt' from singleton 'Samples.Lights.Controller'.
 ---> System.InvalidOperationException: Cannot consume scoped service 'Microsoft.Teams.AI.Models.OpenAI.OpenAIChatPrompt' from singleton 'Samples.Lights.Controller'.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.VisitCallSite(ServiceCallSite callSite, CallSiteValidatorState argument)
[...]

PR #127, opened 2025-09-15
[Bug] Add singleton factory method for scoped prompt

MehakBindra:
Issue : https://github.com/microsoft/teams.net/issues/109

Problem:
In `Samples.Lights.`
`Controller` should be registered as a singleton.
`LightsPrompt` is a scoped service because it varies per request (it maintains state that is accessed via request context).
ASP.NET Core DI forbids injecting a scoped service into a singleton.

Solution: 
A singleton factory is added that creates a new scope and resolves OpenAIChatPrompt from that scope:
```
 var scope = serviceProvider.CreateScope();
 return scope.ServiceProvider.GetRequiredService<OpenAIChatPrompt>();
```
This singleton factory is injected into Controller

### DIFF
--- a/Libraries/Microsoft.Teams.Extensions/Microsoft.Teams.Extensions.Hosting/Microsoft.Teams.AI.Models.OpenAI.Extensions/ServiceCollection.cs
+++ b/Libraries/Microsoft.Teams.Extensions/Microsoft.Teams.Extensions.Hosting/Microsoft.Teams.AI.Models.OpenAI.Extensions/ServiceCollection.cs
@@ -73,6 +73,20 @@ public static class ServiceCollectionExtensions
             return OpenAIChatPrompt.From(model, value, (options ?? new()).WithLogger(logger));
         });
 
-        return collection.AddScoped<IChatPrompt>(provider => provider.GetRequiredService<OpenAIChatPrompt>());
+        collection.AddScoped<IChatPrompt>(provider => provider.GetRequiredService<OpenAIChatPrompt>());
+
+        // Add a singleton factory for creating scoped prompts 
+        // required when added as dependency to singleton controllers
+        collection.AddSingleton<Func<OpenAIChatPrompt>>(provider =>
+        {
+            var serviceProvider = provider;
+            return () =>
+            {
+                var scope = serviceProvider.CreateScope();
+                return scope.ServiceProvider.GetRequiredService<OpenAIChatPrompt>();
+            };
+        });
+
+        return collection;
     }
 }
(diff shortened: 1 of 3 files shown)


Jesperholmbergmsft (approved):

heyitsaamir (dismissed):

Samples/Samples.Lights/Controller.cs
```diff
@@ -9,7 +9,7 @@
 namespace Samples.Lights;
 
 [TeamsController]
-public class Controller(OpenAIChatPrompt _prompt)
+public class Controller(Func<OpenAIChatPrompt> _promptFactory)
```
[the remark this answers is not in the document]
MehakBindra: private field for dependency injected objects

Jesperholmbergmsft (approved):

Libraries/Microsoft.Teams.Extensions/Microsoft.Teams.Extensions.Hosting/Microsoft.Teams.AI.Models.OpenAI.Extensions/ServiceCollection.cs
```diff
@@ -73,6 +73,20 @@ public static IServiceCollection AddOpenAI<T>(this IServiceCollection collection
...
+        collection.AddScoped<IChatPrompt>(provider => provider.GetRequiredService<OpenAIChatPrompt>());
+
+        // Add a singleton factory for creating scoped prompts 
+        // required when added as dependency to singleton controllers
+        collection.AddSingleton<Func<OpenAIChatPrompt>>(provider =>
+        {
+            var serviceProvider = provider;
+            return () =>
+            {
+                var scope = serviceProvider.CreateScope();
```
singhk97, line 85: the overall approach seems right but is it necessary to create a scope. Wouldn't just getting the service from the provider itself resolve it in the given request scope?

singhk97: scopes that are created would have to be [resolved appropriately](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.iservicescopefactory.createscope?view=net-9.0-pp#returns), this makes it a bit harder to reason about what's going on

MehakBindra: System.InvalidOperationException: Cannot resolve scoped service 'Microsoft.Teams.AI.Models.OpenAI.OpenAIChatPrompt' from root provider.

PR #131, opened 2025-09-16
[Bug] Use a separate client for token fetches

singhk97:
- Now uses a separate client for token fetches. This issue also existed beforehand when trying to refresh a token. 
- Allow passing in a custom token factory

### DIFF
--- a/Tests/Microsoft.Teams.Apps.Tests/AppTests.cs
+++ b/Tests/Microsoft.Teams.Apps.Tests/AppTests.cs
@@ -188,4 +188,30 @@ public class AppTests
         api.Verify(api => api.Bots.Token.GetAsync(It.IsAny<IHttpCredentials>(), It.IsAny<IHttpClient>()), Times.Never);
         Assert.Null(app.Token);
     }
+
+    [Fact]
+    public void Test_App_Client_CustomTokenFactory()
+    {
+        // arrange
+        var client = new Mock<Common.Http.HttpClient>() { CallBase = true };
+        var tokenFactoryInvoked = false;
+        IHttpClientOptions.HttpTokenFactory tokenFactory = () =>
+        {
+            tokenFactoryInvoked = true;
+            return null;
+        };
+        client.Object.Options.TokenFactory = tokenFactory;
+        var options = new AppOptions()
+        {
+            Client = client.Object,
+            Credentials = null,
+        };
+        var app = new App(options);
+
+        // act
+        client.Object.Options.TokenFactory();
+
+        // assert
+        Assert.True(tokenFactoryInvoked);
+    }
 }


heyitsaamir (dismissed):

singhk97 (commented):
Added a few comments

Libraries/Microsoft.Teams.Apps/App.cs
```diff
@@ -60,25 +61,26 @@ public App(AppOptions? options = null)
         OAuth = options?.OAuth ?? new OAuthSettings();
         Provider = options?.Provider;
 
+        TokenClient = new Common.Http.HttpClient();
```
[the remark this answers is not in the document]
heyitsaamir: Ya you should.

OR, you could just provide the overridable tokenfactory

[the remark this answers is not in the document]
singhk97: we already support an overridable token factory. to make things less confusing i think we should keep the token client internal and point devs to overriding the token factory for custom implementation

Libraries/Microsoft.Teams.Api/Clients/ApiClient.cs
```diff
@@ -7,10 +7,10 @@ namespace Microsoft.Teams.Api.Clients;
 
 public class ApiClient : Client
 {
-    public readonly string ServiceUrl;
-    public readonly BotClient Bots;
-    public readonly ConversationClient Conversations;
-    public readonly UserClient Users;
+    public virtual string ServiceUrl { get; }
```
singhk97, line 10: Makes the API client mockable

Libraries/Microsoft.Teams.Api/Clients/BotClient.cs
```diff
@@ -7,9 +7,14 @@ namespace Microsoft.Teams.Api.Clients;
 
 public class BotClient : Client
 {
-    public BotTokenClient Token { get; }
+    public virtual BotTokenClient Token { get; }
```
singhk97, line 10: Makes the BotClient Mockable

Libraries/Microsoft.Teams.Api/Clients/BotTokenClient.cs
```diff
@@ -7,6 +7,14 @@ namespace Microsoft.Teams.Api.Clients;
 
 public class BotTokenClient : Client
 {
+    public static readonly string BotScope = "https://api.botframework.com/.default";
+    public static readonly string GraphScope = "https://graph.microsoft.com/.default";
+
+    public BotTokenClient() : this(default)
```
singhk97, line 13: Makes the BotTokenClient Mockable

Libraries/Microsoft.Teams.Apps/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.Teams.Apps.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100fda1d1b3a8f1e375020aeedb26e512a71e0cdd9c854e4c87e181c1152967e79b5ab815e18e0af0f036e06d3dc0811fdb7251bce2fbec8b9492d7e8bc6a68e6bd3ca28c4c4a30b65c5915268b56644cd4b8e870a63baa7709f1d5a445c8a381d7b3ab3f116410081ec9f6dde59785edf01b689d14aeda42ae4e33139bf1ff92b0")]
```
singhk97, line 3: Since Microsoft.Teams.Apps is strong name signed, exposing its internals is only allowed to strong name signed packages. In this case the test project is strong name signed w/ the same public key. 

Think of the public key as an extension of the package name. Nothing to do with authentication.

MehakBindra (approved):

PR #133, opened 2025-09-16
[fix] When client id is empty, skip auth correctly

MehakBindra:
Currently, if ClientID is empty, we do not register the auth policy or scheme, resulting in Policy not found errors at the endpoint.

We now register the policy but bypass authentication if clientID is empty


singhk97:
Nice fix!

PR #150, opened 2025-09-19
Update Cards and do not set nulls for unset values when serializing

heyitsaamir:
We had found some issues with the generated version that we had fixed in our code. Now the generated code took those suggestions and fixed it up. So this is reintegrating it back.

Tested using an app on Teams, and unit tests.

Main changes:
1. MsTeams `msTeams` -> Msteams `msteams`
2. Built in Serialize and Deserialize methods
3. Union -> IUnion

The other major change this PR has is that in responses to /api/messages, we now ignore unset values. We by default didn't have this behavior so it would set all those values to nulls. With this it ignores them. This matches the current behavior with python and typscript.

#### PR Dependency Tree

* **PR #150** 👈

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)


singhk97 (dismissed):

heyitsaamir (commented):
~Do not push - found an issue with this~

All good now

singhk97 (approved):
lgtm

aacebo (approved):

PR #151, opened 2025-09-19
Add overload for `.Use` middleware

singhk97:
Developers aren't forced to `return null` in the handlers if there's nothing to return.

> Changes added to graph samples are used in the docs snippets

### DIFF
--- a/Libraries/Microsoft.Teams.Apps/AppRouting.cs
+++ b/Libraries/Microsoft.Teams.Apps/AppRouting.cs
@@ -185,7 +185,12 @@ public partial class App
     /// <returns></returns>
     public App Use(Func<IContext<IActivity>, Task<object?>> handler)
     {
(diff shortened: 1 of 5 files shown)


Libraries/Microsoft.Teams.Apps/AppRouting.cs
```diff
@@ -193,4 +193,18 @@ public App Use(Func<IContext<IActivity>, Task<object?>> handler)
...
+    /// Register a middleware.
+    /// </summary>
+    /// <param name="handler">Callback to invoke.</param>
+    /// <returns></returns>
+    public App Use(Func<IContext<IActivity>, Task> handler)
+    {
+        return Use(async (context) =>
+        {
+            await handler(context);
+            return null;
```
heyitsaamir, line 207: Should this return null IFF handler returns nothing?

PR #153, opened 2025-09-19
Showcase both minimal and controller in samples

heyitsaamir:
Updates dialog, MEs, and Card samples to showcase both Controller and Minimal apis.
Need to update Docs, but I tested and both work.


lilyydu (dismissed):

lilyydu (approved):

singhk97 (approved):
what's the motivation behind not extracting duplicate logic into a private method?

Samples/Samples.Cards/appsettings.Development.json
```diff
@@ -12,5 +12,6 @@
   "Teams": {
     "ClientId": "",
     "ClientSecret": ""
-  }
+  },
+  "USE_MINIMAL_API": false
```
aacebo, line 16: nit: change to `UseMinimalApi` to match pascal naming conventions of C# appsettings

aacebo:
this may be less confusing for devs if we just create a new sample for minimal, wondering what the teams thoughts are on that. Possibly in a subdirectory `Samples/Samples.Minimal`, where the controller samples stay where they are but we move and create minimal samples in that sub directory.

Samples/Samples.Cards/Program.cs
```diff
@@ -22,7 +25,23 @@ public static void Main(string[] args)
         // Bind the application to localhost:3978 when run with `dotnet run`
         builder.WebHost.UseUrls("http://localhost:3978");
         builder.Services.AddOpenApi();
-        builder.Services.AddTransient<Controller>();
+
+        // Use environment variable to choose between Controller and Minimal API approaches
+        var useMinimalApi = builder.Configuration.GetValue<bool>("USE_MINIMAL_API", false);
```
rido-min, line 30: I think the samples should focus explaining one concept, I would add another sample instead of adding a config to select one or the other.

Samples/Samples.Cards/Program.cs
```diff
@@ -33,93 +52,102 @@ public static void Main(string[] args)
...
-            log.Info($"[MESSAGE] Received: {SanitizeForLog(activity.Text)}");
-            log.Info($"[MESSAGE] From: {SanitizeForLog(activity.From?.Name ?? "unknown")}");
+            var activity = context.Activity;
+
+            context.Log.Info($"[MESSAGE] Received: {SanitizeForLog(activity.Text)}");
+            context.Log.Info($"[MESSAGE] From: {SanitizeForLog(activity.From?.Name ?? "unknown")}");
 
             var text = activity.Text?.ToLowerInvariant() ?? "";
 
             if (text.Contains("card"))
```
rido-min, line 77: why not using routing, instead a big IF/Switch?

Samples/Samples.Cards/Program.cs
```diff
@@ -165,351 +193,495 @@ public async Task<ActionResponse> OnCardAction([Context] AdaptiveCards.ActionAct
...
                 }
-            };
-        }
+            },
+            Actions = new List<Microsoft.Teams.Cards.Action>
+            {
+                new ExecuteAction
+                {
+                    Title = "Save",
+                    Data = new Union<string, SubmitActionData>(new SubmitActionData { NonSchemaProperties = new Dictionary<string, object?> { { "action", "save_profile" } } }),
```
rido-min, line 333: Union is an advanced concept in C# (I'm still learning).. is really required for this sample?

Samples/Samples.Cards/Program.cs
```diff
@@ -165,351 +193,495 @@ public async Task<ActionResponse> OnCardAction([Context] AdaptiveCards.ActionAct
...
+                    Style = ActionStyle.Positive
+                }
+            }
+        };
+    }
+
+    private static AdaptiveCard CreateCardFromJson()
+    {
+        // JSON similar to the Python create_model_validate_card example
+        var cardJson = @"{
```
rido-min, line 433: nit: use the new JSON support with `"""`

Samples/Samples.Cards/Program.cs
```diff
@@ -165,351 +193,495 @@ public async Task<ActionResponse> OnCardAction([Context] AdaptiveCards.ActionAct
...
-                ],
-                ""version"": ""1.5"",
-                ""schema"": ""http://adaptivecards.io/schemas/adaptive-card.json""
-            }";
+            log.Info($"[MESSAGE] Received: {SanitizeForLog(activity.Text)}");
+            log.Info($"[MESSAGE] From: {SanitizeForLog(activity.From?.Name ?? "unknown")}");
+
+            var text = activity.Text?.ToLowerInvariant() ?? "";
+
+            if (text.Contains("card"))
```
rido-min, line 549: same as before.. can we use routing instead of this big IF?

Samples/Samples.Dialogs/appsettings.Development.json
```diff
@@ -13,5 +13,6 @@
     "ClientId": "",
     "ClientSecret": ""
   },
-  "BotEndpoint": ""
+  "BotEndpoint": "",
+  "USE_MINIMAL_API": false
```
rido-min, line 17: dont think this is the place, but if we decided to keep it, move the value to the main appSettings.json

Samples/Samples.MessageExtensions/Program.cs
rido-min, line 1: do we need a manifest to run this sample?

Samples/Samples.Cards/Program.cs
```diff
@@ -33,93 +52,102 @@ public static void Main(string[] args)
...
 
             // Extract action from the Value property
             string? action = data.TryGetValue("action", out var actionObj) ? actionObj?.ToString() : null;
 
             if (string.IsNullOrEmpty(action))
             {
-                log.Error("[CARD_ACTION] No action specified in card data");
-                return new ActionResponse.Message("No action specified") { StatusCode = 400 };
+                context.Log.Error("[CARD_ACTION] No action specified in card data");
+                return new Microsoft.Teams.Api.AdaptiveCards.ActionResponse.Message("No action specified") { StatusCode = 400 };
```
rido-min, line 148: got this error.. is this expected?

Image

Samples/Samples.Cards/Program.cs
```diff
@@ -165,351 +193,495 @@ public async Task<ActionResponse> OnCardAction([Context] AdaptiveCards.ActionAct
...
+            }
+            else if (text.Contains("json"))
+            {
+                log.Info("[JSON] JSON deserialization card requested");
+                var card = CreateCardFromJson();
+                await client.Send(card);
+            }
+            else if (text.Contains("reply"))
+            {
+                await client.Send("Hello! How can I assist you today?");
```
rido-min, line 587: nit.. this is the typical message generated by GTP.. what is the intent of the reply command? maybe remove it?

rido-min (commented):
I see the value of showing minimal and controllers, but I dont think we should add both options to the same sample, as both options should not coexist in the real world. Hence no need to add a config to select which one.

Instead I think we should have an EchoMinimal, and EchoControllers, and then decide which one we should use in all the other samples (I'd say minimal).

Saying that, I have some minor observations, mostly nits

I've been able to run Cards, but Dialogs and MsgExtensions require a manifest that I cannot find.

PR #160, opened 2025-09-24
`AspNetCorePlugin` refactoring

singhk97:
- Moved `AspNetCorePlugin`s `MessageController` logic into a public `Do(HttpContext httpContext)` method. Main benefit is code reuse for when user's define their own endpoint controller.
- Introduced `AspNetCorePlugin.ExtractActivity(HttpContex httpContex)` and `AspNetCorePlugin.ExtractActivity(HttpContex httpContex)` utility methods.

## Testing
- Ran `Samples.BotBuilder` in Teams
- Ran `Samples.Echo` in Teams
- Added units tests

### DIFF
--- a/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore.BotBuilder/Controllers/MessageController.cs
+++ b/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore.BotBuilder/Controllers/MessageController.cs
@@ -1,15 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System.Text.Json;
-
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Bot.Builder;
 using Microsoft.Bot.Builder.Integration.AspNet.Core;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Teams.Api.Activities;
-using Microsoft.Teams.Api.Auth;
+using Microsoft.Teams.Apps.Events;
 
 namespace Microsoft.Teams.Plugins.AspNetCore.BotBuilder
 {
@@ -34,11 +32,10 @@ namespace Microsoft.Teams.Plugins.AspNetCore.BotBuilder
         [HttpPost("/api/messages")]
         public async Task<IResult> PostAsync()
         {
-            HttpContext.Request.EnableBuffering();
-            var body = await new StreamReader(Request.Body).ReadToEndAsync();
-            Activity? activity = JsonSerializer.Deserialize<Activity>(body);
-            HttpContext.Request.Body.Position = 0;
-
+            // Extract the activity from the incoming request
+            // Activity has to be extracted before calling ProcessAsync
+            // because ProcessAsync will consume the request body stream
+            Activity? activity = await _plugin.ExtractActivity(HttpContext.Request);
             if (activity == null)
             {
                 return Results.BadRequest("Missing activity");
@@ -53,10 +50,9 @@ namespace Microsoft.Teams.Plugins.AspNetCore.BotBuilder
                 return Results.Empty;
             }
 
-            // Fallback logic
-            var authHeader = HttpContext.Request.Headers.Authorization.FirstOrDefault() ?? throw new UnauthorizedAccessException();
-            var token = new JsonWebToken(authHeader.Replace("Bearer ", ""));
-            var res = await _plugin.Do(new()
+            // Fallback logic use the plugin to process the activity
+            var token = _plugin.ExtractToken(HttpContext.Request);
+            var res = await _plugin.Do(new ActivityEvent()
             {
                 Token = token,
                 Activity = activity,
--- a/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore.BotBuilder/Controllers/MessageController.cs
+++ b/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore.BotBuilder/Controllers/MessageController.cs
@@ -6,8 +6,6 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Bot.Builder;
 using Microsoft.Bot.Builder.Integration.AspNet.Core;
 using Microsoft.Extensions.Hosting;
-using Microsoft.Teams.Api.Activities;
-using Microsoft.Teams.Apps.Events;
 
 namespace Microsoft.Teams.Plugins.AspNetCore.BotBuilder
 {
@@ -32,14 +30,8 @@ namespace Microsoft.Teams.Plugins.AspNetCore.BotBuilder
         [HttpPost("/api/messages")]
         public async Task<IResult> PostAsync()
         {
-            // Extract the activity from the incoming request
-            // Activity has to be extracted before calling ProcessAsync
-            // because ProcessAsync will consume the request body stream
-            Activity? activity = await _plugin.ExtractActivity(HttpContext.Request);
-            if (activity == null)
-            {
-                return Results.BadRequest("Missing activity");
-            }
+            // Enable buffering so that the request can be read by the adapter and the plugin
+            HttpContext.Request.EnableBuffering();
 
             // Delegate the processing of the HTTP POST to the adapter.
             // The adapter will invoke the bot.
@@ -51,15 +43,7 @@ namespace Microsoft.Teams.Plugins.AspNetCore.BotBuilder
             }
 
             // Fallback logic use the plugin to process the activity
-            var token = _plugin.ExtractToken(HttpContext.Request);
-            var res = await _plugin.Do(new ActivityEvent()
-            {
-                Token = token,
-                Activity = activity,
-                Services = HttpContext.RequestServices
-            }, _lifetime.ApplicationStopping);
-
-            return Results.Json(res.Body, statusCode: (int)res.Status);
+            return await _plugin.Do(HttpContext, _lifetime.ApplicationStopping);
         }
     }
 }
--- a/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
+++ b/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Text.Json;
+
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Teams.Api.Activities;
 using Microsoft.Teams.Api.Auth;
 using Microsoft.Teams.Api.Clients;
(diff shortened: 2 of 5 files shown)


Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -139,4 +147,85 @@ await Events(
...
+            };
+
+            foreach (var pair in httpContext.Items)
+            {
+                var key = pair.Key.ToString();
+
+                if (key is null) continue;
+
+                data[key] = pair.Value;
+            }
```
heyitsaamir, line 176: hmm, we don't do this in other SDKs. We should right? What is essential for propagation @singhk97 ?

heyitsaamir (dismissed):

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -139,4 +147,85 @@ await Events(
...
+        return new JsonWebToken(authHeader.Replace("Bearer ", ""));
+    }
+
+    public async Task<Activity?> ExtractActivity(Microsoft.AspNetCore.Http.HttpRequest httpRequest)
+    {
+        // Fallback logic
+        httpRequest.EnableBuffering();
+        var body = await new StreamReader(httpRequest.Body).ReadToEndAsync();
+        Activity? activity = JsonSerializer.Deserialize<Activity>(body);
+        httpRequest.Body.Position = 0;
```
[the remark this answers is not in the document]
heyitsaamir: interesting. why?

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
aacebo: I believe I originally did this in the controller to support a PA ask about adding their own context data in a key value pair, this comes off the request so if node/python have a concept like this we should def carry it over. I don't believe express does (may be mistaken) but the JS implementation of the context is already setup to access key/value pairs.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore.BotBuilder/Controllers/MessageController.cs
```diff
@@ -53,10 +50,9 @@ public async Task<IResult> PostAsync()
                 return Results.Empty;
             }
 
-            // Fallback logic
-            var authHeader = HttpContext.Request.Headers.Authorization.FirstOrDefault() ?? throw new UnauthorizedAccessException();
-            var token = new JsonWebToken(authHeader.Replace("Bearer ", ""));
-            var res = await _plugin.Do(new()
+            // Fallback logic use the plugin to process the activity
```
aacebo, line 53: Q: what does this fallback logic note mean? From what I can see this isn't fallback logic it is the main logical path to use the plugin to process the activity.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore.BotBuilder/Controllers/MessageController.cs
```diff
@@ -53,10 +50,9 @@ public async Task<IResult> PostAsync()
...
+            var token = _plugin.ExtractToken(HttpContext.Request);
+            var res = await _plugin.Do(new ActivityEvent()
```
aacebo, line 55: since we now have the Do which takes in the request should we use that here to minimize code?

aacebo (commented):
minor comments but overall implementation looks solid

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/Controllers/MessageController.cs
```diff
@@ -26,47 +24,8 @@ public MessageController(AspNetCorePlugin plugin, IHostApplicationLifetime lifet
...
-
-        return Results.Json(
-            res.Body,
-            new System.Text.Json.JsonSerializerOptions
-            {
-                DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull
-            },
-            contentType: null,
-            statusCode: (int)res.Status);
+        return await _plugin.Do(HttpContext, _lifetime.ApplicationStopping);
```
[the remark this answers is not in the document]
rido-min: >  verbosity of the minimal implementation

LoL

let's chat, I'm not sure I follow this question. If it's a controller it's not a minimal API.. no?

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -27,6 +30,11 @@ public partial class AspNetCorePlugin : ISenderPlugin, IAspNetCorePlugin
 
     public event EventFunction Events;
 
+    private static readonly JsonSerializerOptions _jsonSerializerOptions = new()
```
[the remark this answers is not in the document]
rido-min: mhhh this is tricky, as with different options we could break some scenarios when not using the expected converters and such..

To follow the Open/Closed principle I'd say the options should be internal, and if we need to support a different serializer let's delegate the complete serialization process instead of just the options.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore.BotBuilder/Controllers/MessageController.cs
singhk97: If the BF adapter doesn't return an invoke response then the control flow will fallback to Teams sdk app.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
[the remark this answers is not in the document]
singhk97: There's no simple way to for thr `Utf8JsonReader` to consume a `Stream` (req body) and so it's more performant to keep it this way imo.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/Controllers/MessageController.cs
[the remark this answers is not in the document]
singhk97: Will open a separate PR for this refactor since it needs more thorough testing to ensure things don't break.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -139,4 +147,91 @@ await Events(
...
+    {
+        httpRequest.EnableBuffering();
+
+        if (httpRequest.Body.CanSeek)
+        {
+            // reset the stream position to the beginning in case it was read before
+            httpRequest.Body.Position = 0;
+        }
+
+        var body = await new StreamReader(httpRequest.Body).ReadToEndAsync();
```
rido-min, line 231: StreamReader is Disposable missing one `using` like

```cs
using StreamReader sr = new(httpRquest.Body);
var body = await sr.ReadToEndAsync();
```

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -139,4 +147,91 @@ await Events(
...
+        }
+    }
+
+    public JsonWebToken ExtractToken(Microsoft.AspNetCore.Http.HttpRequest httpRequest)
+    {
+        var authHeader = httpRequest.Headers.Authorization.FirstOrDefault() ?? throw new UnauthorizedAccessException();
+        return new JsonWebToken(authHeader.Replace("Bearer ", ""));
+    }
+
+    public async Task<Activity?> ExtractActivity(Microsoft.AspNetCore.Http.HttpRequest httpRequest)
```
rido-min, line 221: nit. I'd call this method with a name explaining what is doing, since Deserializing involves some validation.. maybe `ParseActivity`? 

imho _extract_ means just grabbing specific parts.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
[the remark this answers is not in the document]
rido-min: leaving the Utf8JsonReader comment aside (you are right here). 

Is there any case for allow to rewind the request stream? If not, I would not set the position to 0.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -27,6 +30,11 @@ public partial class AspNetCorePlugin : ISenderPlugin, IAspNetCorePlugin
 
     public event EventFunction Events;
 
+    private static readonly JsonSerializerOptions _jsonSerializerOptions = new()
+    {
+        DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull
```
rido-min, line 35: nit: Add using instead of long namespaces?

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -139,4 +147,91 @@ await Events(
             return new Response(System.Net.HttpStatusCode.InternalServerError, ex.ToString());
         }
     }
+
+    public async Task<IResult> Do(HttpContext httpContext, CancellationToken cancellationToken = default)
```
rido-min, line 151: DoAsync? 

Wondering if we could pass the cancellation token

rido-min (approved):
nothing blocking, just minor comments

Tests/Microsoft.Teams.Plugins.AspNetCore.Tests/AspNetCorePluginTests.cs
```diff
@@ -0,0 +1,183 @@
...
+
+    [Fact]
+    public async Task Test_Do_Http_CallsExtractTokenAndActivity_AndCallsCoreDo()
+    {
+        // Arrange
+        var activity = CreateMessageActivity();
+        var coreResponse = new Response(HttpStatusCode.Accepted, new { ok = true });
+        var eventsCalled = new List<string>();
+
+        EventFunction events = async (plugin, name, payload, ct) =>
```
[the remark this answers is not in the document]
rido-min: curious why you don't get warnings when not adding the _Async_ suffix to async methods? I believe that rule is configured by default in all the CodeAnalysis rules.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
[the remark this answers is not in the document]
singhk97: No specific reason, I've removed it. What's the reason though? Does it have to do with the object not being properly disposed?

PR #167, opened 2025-09-30
Handle multiple "Message" entities on a Message Activity

heyitsaamir:
Add multiple message entities in the same Activity. Without this, if you called addAIGenerated or addCitations twice, the service would throw.

#### PR Dependency Tree

* **PR #167** 👈
  * **PR #168**

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)


Libraries/Microsoft.Teams.Api/Activities/Activity.cs
```diff
@@ -287,18 +287,58 @@ public virtual Activity AddEntity(params IEntity[] entities)
...
+        }
+
+        Entities.Add(newEntity);
+        return this;
+    }
+
+    /// <summary>
+    /// ensures a single root level message entity exists
+    /// </summary>
+    private IMessageEntity EnsureSingleRootLevelMessageEntity()
```
aacebo, line 308: Nit: since you are returning the entity maybe `GetRootLevelMessageEntity` would be easier to understand at a glance.

heyitsaamir:
Talked to @aacebo offline - There's some confusion if we need OMessageEntity and MessageEntity. He'll take care of it in a follow-up

lilyydu (approved):

aacebo (approved):
thanks for fixing the warnings!

lilyydu (approved):

PR #169, opened 2025-10-01
Make Conversation.Type optional; improve JSON handling

rido-min:
The `Conversation.Type` property was changed from required to optional (`ConversationType?`) to allow null values, improving flexibility.

Removed default initialization of `From`, `Recipient`, and `Conversation` in `Context<TActivity>.Send`, delegating responsibility to the caller.

Updated `HttpClient` to use `StringContent` for JSON serialization and switched deserialization to `Dictionary<string, JsonElement>` for better type safety.

Added `ChannelId` to `AspNetCorePlugin` activities to ensure channel information is included.

Enhanced test coverage in `MessageActivityTests` with new tests for serialization and deserialization of `MessageActivity` objects with optional `Conversation.Type`.

Added `MessageActivity_webChat.json` to support new test cases.


Libraries/Microsoft.Teams.Common/Http/HttpClient.cs
```diff
@@ -89,10 +89,11 @@ protected HttpRequestMessage CreateRequest(IHttpRequest request)
             }
             else
             {
-                httpRequest.Content = JsonContent.Create(request.Body, options: new JsonSerializerOptions()
+                string body = JsonSerializer.Serialize(request.Body, new JsonSerializerOptions()
```
rido-min, line 92: Looks like ABS requires an explicit UTF8 encoding (or something like that), and I was not able to make it work with `JsonContent`, changing to `StringContent` made the trick

Libraries/Microsoft.Teams.Common/Http/HttpClient.cs
```diff
@@ -172,7 +173,7 @@ private async Task<object> ParseErrorBody(HttpResponseMessage response)
 
         try
         {
-            var bodyAsJson = JsonSerializer.Deserialize<Dictionary<string, object>>(content);
+            var bodyAsJson = JsonSerializer.Deserialize<Dictionary<string, JsonElement>>(content);
```
rido-min, line 176: this is unrelated to this PR, but It took me awhile to figure out why the error messages were not printed out, this happend when the json has a nested object, I believe that with JsonElement we won't loose the values..

And TBH, I would return just the string, since we cannot guarantee the server will always return errors as json.

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -90,6 +90,7 @@ public async Task<TActivity> Send<TActivity>(TActivity activity, Api.Conversatio
         activity.Conversation = reference.Conversation;
         activity.From = reference.Bot;
         activity.Recipient = reference.User;
+        activity.ChannelId = reference.ChannelId;
```
rido-min, line 93: interestingly we were passing teams to webchat :)

Tests/Microsoft.Teams.Api.Tests/Activities/Message/MessageActivityTests.cs
```diff
@@ -220,4 +221,56 @@ public void JsonDeserialize_Derived()
...
+        string exptected = File.ReadAllText(@"../../../Json/Activity/Message/MessageActivity_webChat.json");
+        Assert.Equal(exptected, json);
+    }
+
+    [Fact]
+    public void JsonDeserialize_WebChat_AllowsEmptyConversationType()
+    {
+        var json = File.ReadAllText(@"../../../Json/Activity/Message/MessageActivity_webChat.json");
+        var activity = JsonSerializer.Deserialize<MessageActivity>(json);
+        Assert.NotNull(activity);
```
rido-min, line 259: I prefer explicit asserts, rather than Assert.Equivalent, especially when dealing with null/empty values. 

For consistency we can update all the other Assertions in this file (or even this project)

rido-min:
> I think this change makes sense, and we should bring it to other languages too.

In JS I've not seen this behavior
 
> Does this impact the ergonomics of "send"? 

I dont think so, 

> I'm assuming you tested with some of the existing samples too?

I tried Echo and Cards, I believe all the other tamples are Teams specific

Libraries/Microsoft.Teams.Common/Http/HttpClient.cs
singhk97: Creating a new `JsonSerializerOptions` will cause performance overheards since it generates and caches metadata internally. The [guidance](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/configure-options#reuse-jsonserializeroptions-instances) is to reuse the same instance (it's thread-safe btw)

singhk97 (changes requested):
lgtm ... requested one change

rido-min:
> lgtm ... requested one change

updated

Libraries/Microsoft.Teams.Common/Http/HttpClient.cs
```diff
@@ -18,13 +18,18 @@ public interface IHttpClient : IDisposable
     public Task<IHttpResponse<TResponseBody>> SendAsync<TResponseBody>(IHttpRequest request, CancellationToken cancellationToken = default);
 }
 
-public class HttpClient : IHttpClient
+public class HttpClient : IHttpClient
```
corinagum, line 21: nit: run lint?

corinagum (approved):
Thank you @rido-min

Issue #172, opened 2025-10-05
Token refresh fails after idle time due to missing scope in OAuth request


LZIde:
**Problem Summary**
After some idle time, the Microsoft Teams bot fails to refresh its authentication token. The initial bot calls work fine, but subsequent requests after the token expires produce an OAuth error due to a missing 'scope' parameter in the token request.

**Detailed Stack Trace and Error**
```
[ERROR] Teams {
  "error": "invalid_request",
  "error_description": "AADSTS90014: The required field 'scope' is missing from the credential. Ensure that you have all the necessary parameters for the login request. Trace ID: ... Correlation ID: ... Timestamp: 2025-10-05 12:25:02Z",
  "error_codes": [90014],
  "timestamp": "2025-10-05 12:25:02Z",
  "trace_id": "...",
  "correlation_id": "...",
  "error_uri": "https://login.microsoftonline.com/error?code=90014"
}
[ERROR] Teams https://login.microsoftonline.com/<tenant>/oauth2/v2.0/token
[ERROR] Teams grant_type=client_credentials&client_id=<client_id>&client_secret=<client_secret>&scope=
```

**Root Cause**
- The scope parameter is empty when requesting a new token after expiration.
- In `App.cs`, the code path for initial token fetch and token refresh is inconsistent:
    - Initial fetch uses `Api.Bots.Token.GetAsync(Credentials, TokenClient)` (works)
    - Refresh uses `Credentials.Resolve(TokenClient, [.. Token.Scopes])`, but `Token.Scopes` is empty/null after parsing the expired token, resulting in `scope=` being empty.
- This causes the OAuth request to fail with error code AADSTS90014.

**How to Reproduce**
1. Start the bot and make requests—authentication works initially.
2. Allow the bot to idle until its token expires.
3. Make a new request—the token refresh logic triggers and the error occurs.

**Expected Behavior**
- Token refresh should use the same logic as initial token fetch, or ensure that a valid scope is always supplied.

**Suggested Fix**
- Ensure that a default scope (e.g., `https://api.botframework.com/.default`) is always provided when refreshing tokens.
- Update the logic in `App.cs` to avoid passing an empty scope array to the OAuth request.

**Relevant Code Location**
See constructor in `Libraries/Microsoft.Teams.Apps/App.cs`, where token refresh logic is implemented.

bigbearzhu:
Thanks for fixing the bug, we are hitting the same issue at the moment. Do you have any idea when this is going to be released?

[...]

PR #175, opened 2025-10-09
Add default scopes for token refresh

mauricehennig:
Fixes https://github.com/microsoft/teams.net/issues/172 by checking if the auth server echoed back all scopes. If not, default scope is set


bigbearzhu:
Hope to get this fix released in nuget soon, 2.0.3 doesn't seem to have it yet.

PR #191, opened 2025-10-22
[Fix] streaming: do not reset timeout for each chunk emit

MehakBindra:
Issue : https://github.com/microsoft/teams.ts/issues/374

Main change::

- In emit, instead of cancelling and scheduling a new timeout,
   - flush immediately if no timeout pending,
   - else push to queue and wait.
- Set retry to 500 ms to avoid API quota exceeded errors

EDIT 
- Changed the behavior of lock: Instead of blocking, it now tries to acquire the lock and immediately returns if another flush is running.
- `WaitForIdAndQueueAsync` : On close, donot wait forever for id to be set or queue to be empty. Leads to hanging

https://github.com/user-attachments/assets/dee8e5d7-771f-4a20-bc7b-8eda133a39be

Devtools: Streaming always starts before the full response is received (check with logs, see attached video with the green coloured log).
Teams: Even though the first chunk is emitted immediately, (sometimes) by the time the stream starts on Teams, the full response is ready. In the third msg ?, we can see the stream starts a slightly before the full response is logged.


Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.Stream.cs
```diff
@@ -34,44 +48,44 @@ public class Stream : IStreamer
         private DateTime? _closedAt;
         private int _count = 0;
         private MessageActivity? _result;
-        private readonly SemaphoreSlim _lock = new(1, 1);
+        private bool _flushing = false;
```
heyitsaamir: NIT: is it common for private variables to have _ in C#? I know that's common in like C++...

aacebo (changes requested):

PR #210, opened 2025-11-21
Fix App Send Targeted Messages by updating ConversationReference User

ShanmathiMayuramKrithivasan:
## Problem
`App.Send()` cannot send targeted messages because it builds `ConversationReference`, dropping the required `User` field that specifies the recipient id for targeted messages.

## Fix
- Set `ConversationReference.User = activity.Recipient` when `isTargeted=true` in `App.Send()`
- This ensures the sender plugin can properly set the outgoing activity's recipient
- Maintains backward compatibility for non-targeted messages (`User` remains `null`)

### DIFF
--- a/Libraries/Microsoft.Teams.Apps/App.cs
+++ b/Libraries/Microsoft.Teams.Apps/App.cs
@@ -195,7 +195,7 @@ public partial class App
 
         if (isTargeted && activity.Recipient is null)
         {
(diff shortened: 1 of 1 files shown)


singhk97 (approved):
lgtm

singhk97 (dismissed):

singhk97 (approved):

PR #242, opened 2025-12-12
[do not merge] User assigned managed identity

heyitsaamir:
Tested with a hard-coded value, and it seems to work. Need to clean this up


heyitsaamir:
Merged in #244

PR #244, opened 2025-12-13
Add UMI support

heyitsaamir:
Added UMI support to Core. This is done via `ManagedIdentityOptions` that need to be set. This type of auth is a bit off-band with the rest of MSAL.

* Added UMI support (tested it)
* Add some debug logging to tell us what kind of logs are being used. Unsure if this is the right approach TBH.
* Add some unit tests

### DIFF
--- a/core/src/Microsoft.Bot.Core/Hosting/AddBotApplicationExtensions.cs
+++ b/core/src/Microsoft.Bot.Core/Hosting/AddBotApplicationExtensions.cs
@@ -147,6 +147,7 @@ public static class AddBotApplicationExtensions
 
         services.Configure<MicrosoftIdentityApplicationOptions>(options =>
         {
+            // TODO: Make Instance configurable
             options.Instance = "https://login.microsoftonline.com/";
             options.TenantId = tenantId;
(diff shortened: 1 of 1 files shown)


core/src/Microsoft.Bot.Core/Hosting/AddBotApplicationExtensions.cs
```diff
@@ -187,19 +187,69 @@ private static IServiceCollection ConfigureMSALWithFIC(this IServiceCollection s
...
+        string? umiClientId = isSystemAssigned ? null : (managedIdentityClientId ?? clientId);
+
+        services.Configure<ManagedIdentityOptions>(options =>
+        {
+            options.UserAssignedClientId = umiClientId;
+        });
+
+        services.Configure<MicrosoftIdentityApplicationOptions>(options =>
+        {
+            options.Instance = "https://login.microsoftonline.com/";
```
rido-min: this means we won't support non public clouds when not using MSAL config scheme.. 

For parity, this is something we could add to our CoreConfig...  

@heyitsaamir any thoughts?

PR #262, opened 2026-01-09
Fix DevTools not logging outgoing messages

### LINKED ISSUE #154: Devtools does not log outgoing messages for C#
Image

Incoming messages show up, but not outgoing.

rajan-chari:
## Summary
Fixes issue #154 where DevTools was only displaying incoming messages but not outgoing messages, making it difficult to debug bot conversations during development.

## Problem
The DevTools UI was correctly showing incoming messages but outgoing messages were not being captured or displayed. This significantly impacted the developer experience when debugging bots.

## Root Cause
The event routing logic in `AppEvents.cs` used `.Equals()` to determine which plugins to skip when broadcasting events. The intention was to skip the plugin that sent the message (to avoid notifying it of its own action), but this comparison was incorrectly causing DevToolsPlugin to be skipped as well.

### Core Fix
Replaced `.Equals()` with `ReferenceEquals()` in three event handlers in `Libraries/Microsoft.Teams.Apps/AppEvents.cs`:
- `OnErrorEvent` (line 32)
- `OnActivitySentEvent` (line 55)
- `OnActivityResponseEvent` (line 72)

`ReferenceEquals()` explicitly checks if two references point to the exact same object instance, ensuring only the actual sender plugin is skipped, not observer plugins like DevToolsPlugin.

### Debugging Enhancements
Added detailed debug logging to help diagnose similar issues in the future:
- `AppEvents.cs`: Added logging to track sender, plugin iterations, equality checks, and skip decisions
- `DevToolsPlugin.cs`: Added logging to confirm when OnActivitySent is called and how many WebSocket connections receive the event

## Testing
✅ All 559 existing tests pass  
✅ Build succeeds with no errors or warnings  
✅ No breaking changes to public APIs  

## Verification Steps for Reviewers
1. Run a sample bot with DevTools enabled
2. Send a message from the bot
3. Verify that outgoing messages now appear in the DevTools UI alongside incoming messages

Closes #154

🤖 Generated with Claude Code

### DIFF
--- a/Libraries/Microsoft.Teams.Apps/AppEvents.cs
+++ b/Libraries/Microsoft.Teams.Apps/AppEvents.cs
@@ -29,7 +29,7 @@ public partial class App
 
         foreach (var plugin in Plugins)
         {
-            if (sender.Equals(plugin)) continue;
+            if (ReferenceEquals(sender, plugin)) continue;
             await plugin.OnError(this, sender, @event, cancellationToken);
         }
     }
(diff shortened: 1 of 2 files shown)


rajan-chari:
## Addressed Review Feedback

Thank you for the review. I have addressed all the feedback:

### Changes Made

1. **Removed debug logging**: Removed all temporary debug log statements from both `AppEvents.cs` and `DevToolsPlugin.cs`. Only the original `Logger.Debug()` calls remain, following the existing codebase patterns.

2. **Refactored filtering logic**: Updated all three event handlers to use `.Where()` for explicit filtering instead of `continue` statements inside the loops:
   - `OnErrorEvent`
   - `OnActivitySentEvent`
   - `OnActivityResponseEvent`

### Testing

✅ All 559 tests pass  
✅ Build succeeds with no errors  
✅ Code now follows the existing codebase patterns more closely

The core fix (using `ReferenceEquals()` instead of `.Equals()`) remains unchanged and resolves the issue where DevToolsPlugin was being incorrectly skipped during event notification.

corinagum (approved):

PR #295, opened 2026-01-28
Replace JsonSchema.Net with NJsonSchema

rajan-chari:
## Summary
- Replace `JsonSchema.Net` with `NJsonSchema` (MIT) for JSON schema validation ahead of upcoming JsonSchema.Net license change
- Introduce `IJsonSchema` abstraction interface to decouple the public API from the specific schema library
- Add equivalence tests comparing both libraries to validate behavioral compatibility

## Changes
- **New files:**
  - `IJsonSchema.cs` - Abstraction interface for JSON schema operations
  - `JsonSchemaWrapper.cs` - NJsonSchema-based implementation
  - `JsonSchemaEquivalenceTests.cs` - Tests comparing both libraries
  - (JsonSchema dependency and JsonSchemaEquivalenceTests.cs will be removed once this PR is merged)

- **Modified files:**
  - `Function.cs` - Use `IJsonSchema` instead of `JsonSchema`
  - `ChatPrompt.Chain.cs` / `ChatPrompt.Functions.cs` - Use abstraction
  - `ChatTool.cs` (OpenAI) - Use abstraction
  - `McpClientPlugin.cs` - Use abstraction

## Breaking Change
`IFunction.Parameters` type changes from `JsonSchema?` to `IJsonSchema?`

**User migration required if:**
- Passing `JsonSchema` to `Function` constructor → use `JsonSchemaWrapper.FromJson()` or `JsonSchemaWrapper.FromType()`
- Reading `Parameters` and using `JsonSchema` methods → use `IJsonSchema` methods (`Validate()`, `ToJson()`)

**No changes needed if:**
- Using fluent API with auto-generated schemas: `Function(name, description, handler)`

## Test plan
- [x] All existing tests pass (582 tests)
- [x] New equivalence tests verify both libraries validate identically
- [ ] Manual verification with Samples.Lights

🤖 Generated with [Claude Code](https://claude.ai/code)


Libraries/Microsoft.Teams.AI/Schema/JsonSchemaWrapper.cs
```diff
@@ -0,0 +1,89 @@
...
+            Type = JsonObjectType.String,
+            Description = description
+        };
+        return new JsonSchemaWrapper(schema);
+    }
+
+    /// <inheritdoc/>
+    public JsonSchemaValidationResult Validate(string json)
+    {
+        var errors = _schema.Validate(json);
```
corinagum, line 79: `Function.Invoke` used to validate with `EvaluateAs = SpecVersion.DraftNext` (`Function.cs:85` on main). NJsonSchema's default validator picks a draft from the schema's `$schema` keyword, falling back to Draft-04/06/07 depending on shape. Schemas coming from MCP servers (`McpClientPlugin.cs:221`) and OpenAI tool calls don't declare a draft, so draft-2020-12 keywords (`prefixItems`, `unevaluatedProperties`, `$dynamicRef`) get silently ignored at validation time. Either set the draft explicitly when parsing, or extend the equivalence tests to exercise 2020-12 keywords so this doesn't regress quietly.

Libraries/Microsoft.Teams.AI.Models.OpenAI/Extensions/ChatTool.cs
```diff
@@ -28,7 +24,7 @@ public static ChatTool ToOpenAI(this IFunction function)
         return ChatTool.CreateFunctionTool(
             function.Name,
             function.Description,
-            function.Parameters is null ? null : BinaryData.FromString(JsonSerializer.Serialize(function.Parameters)),
+            function.Parameters is null ? null : BinaryData.FromString(function.Parameters.ToJson()),
```
corinagum, line 27: Was `JsonSerializer.Serialize(function.Parameters)`, now `function.Parameters.ToJson()`. NJsonSchema's `ToJson()` includes `"$schema": "http://json-schema.org/draft-04/schema#"` and may include `definitions`/`x-` extensions. OpenAI tolerates `$schema` but rejects `definitions` in strict mode. There's no test that round-trips `ChatTool.ToOpenAI(...)` output and asserts it's the JSON OpenAI actually accepts. Add one, or post-process `ToJson()` to strip metadata.

Libraries/Microsoft.Teams.AI/Prompts/ChatPrompt/ChatPrompt.Chain.cs
```diff
@@ -17,8 +15,8 @@ public ChatPrompt<TOptions> Chain(IChatPrompt<TOptions> prompt)
         Functions.Add(new Function(
             prompt.Name,
             prompt.Description,
-            new JsonSchemaBuilder().Properties(
-                ("text", new JsonSchemaBuilder().Type(SchemaValueType.String).Description("text to send"))
+            JsonSchemaWrapper.CreateObjectSchema(
+                ("text", JsonSchemaWrapper.String("text to send"), true)
```
corinagum, line 19: The old code passed no `Required(...)` on `text`, so it was optional. New code passes `required: true` here. That's a behavior change, not a swap. Call it out in the PR description as intentional (and confirm it is). If unintentional, drop the `true`.

corinagum (commented):
Two PR-description items worth tightening:

1. The plan to remove `JsonSchema.Net` from the test project and delete `JsonSchemaEquivalenceTests.cs` in a follow-up creates a window where the test assembly depends on a library the production code no longer uses. Either keep both as a permanent regression guard (rename the file to `JsonSchemaRegressionTests.cs`) or strip both in this PR.
2. The "Manual verification with `Samples.Lights`" checkbox is unchecked. For a swap that touches the tool-calling path end-to-end, that should run green before this merges.

Tests/Microsoft.Teams.AI.Tests/JsonSchemaEquivalenceTests.cs
```diff
@@ -0,0 +1,236 @@
...
+using Json.Schema.Generation;
+
+namespace Microsoft.Teams.AI.Tests;
+
+/// <summary>
+/// Equivalence tests comparing JsonSchema.Net with NJsonSchema (via JsonSchemaWrapper).
+/// These tests ensure the migration to NJsonSchema maintains behavioral compatibility.
+/// This file can be removed once the migration is validated and merged.
+/// </summary>
+public class JsonSchemaEquivalenceTests
```
corinagum, line 16: The equivalence suite covers primitives, one POCO, and required-property checks. It doesn't cover any of: arrays of objects, nested objects, enum-typed parameters, nullable refs (`int?`, `Nullable<T>`), `DateTime`, `Guid`, `$ref`, `additionalProperties: false`, or `format`. For a library swap targeting "behavioral compatibility," that's too narrow. Add at minimum array-of-objects and nullable-int. Both surface real differences between the two libraries.

singhk97:
closing since these PRs target now deprecated packages

PR #317, opened 2026-02-10
Add missing cancellation tokens

rido-min:
This pull request introduces comprehensive support for passing `CancellationToken` to activity and command handler registrations throughout the Microsoft Teams app framework. This enhancement allows for improved cancellation and timeout handling in asynchronous operations, making the codebase more robust and responsive. Additionally, a new constructor for `ApiClient` enables explicit control over cancellation tokens, and a local settings file is added for permission configuration.

### Enhanced handler registration with `CancellationToken` support

**Activity handler improvements:**

* Added overloads to `OnActivity` and related methods in `Activity.cs` to accept handler functions with a `CancellationToken` parameter, supporting both generic and non-generic versions, and selector-based registrations. This enables proper cancellation management in all activity handler scenarios. [[1]](diffhunk://#diff-3c8a7200d4052a66f516d17cd3ab173ea1355038fcbaaefc99c72f40e3fa1a61R32-R54) [[2]](diffhunk://#diff-3c8a7200d4052a66f516d17cd3ab173ea1355038fcbaaefc99c72f40e3fa1a61R72-R88) [[3]](diffhunk://#diff-3c8a7200d4052a66f516d17cd3ab173ea1355038fcbaaefc99c72f40e3fa1a61R102-R114) [[4]](diffhunk://#diff-3c8a7200d4052a66f516d17cd3ab173ea1355038fcbaaefc99c72f40e3fa1a61R132-R148) [[5]](diffhunk://#diff-3c8a7200d4052a66f516d17cd3ab173ea1355038fcbaaefc99c72f40e3fa1a61R162-R174) [[6]](diffhunk://#diff-3c8a7200d4052a66f516d17cd3ab173ea1355038fcbaaefc99c72f40e3fa1a61R192-R208) [[7]](diffhunk://#diff-3c8a7200d4052a66f516d17cd3ab173ea1355038fcbaaefc99c72f40e3fa1a61R221-R233)

**Command and result handler improvements:**

* Added `CancellationToken`-aware overloads for `OnCommand` in `CommandActivity.cs` and `OnCommandResult` in `CommandResultActivity.cs`, allowing command handlers to respond to cancellation requests. [[1]](diffhunk://#diff-f10af99e19ce61b4820c61da75c4ced9b7d8c26602bcce2292bf0c69c71f9e50R33-R49) [[2]](diffhunk://#diff-ff06a772342a8a4657771f9c36062da9cbdf4dac199f57fa0bc31df04c3afd9fR33-R49)

[...]

PR #325, opened 2026-02-11
fix transitive package dependency issue + refactor router setup

singhk97:
- Create Router internally with dedicated logger instead of injecting as dependency


rido-min (changes requested):
undo the pin, as the warning was fixed in https://github.com/microsoft/teams.net/pull/317 by updating top level packages

https://github.com/microsoft/teams.net/blob/a14c0afcdf493eb3a29230b5a608ed0b88469314/core/src/Microsoft.Teams.Bot.Core/Microsoft.Teams.Bot.Core.csproj#L18-L25

PR #326, opened 2026-02-12
Add invoke payloads and handlers

MehakBindra:
* Added all invoke activities ( some are commented that could not be tested )
* Added `AllInvokesBot` sample project for adaptive card actions, task modules, file consent, and other invoke scenarios. 
* Added `MessageExtensionBot` sample project.
* Made invoke activity and invoke response strongly typed vi generics.


core/src/Microsoft.Teams.Bot.Apps/Schema/Invokes/MessagePayload.cs
```diff
@@ -0,0 +1,262 @@
...
+    /// </summary>
+    [JsonPropertyName("user")]
+    public User? User { get; set; }
+}
+
+/// <summary>
+/// Message importance levels.
+/// </summary>
+[JsonConverter(typeof(JsonStringEnumConverter<MessagePayloadImportance>))]
+public enum MessagePayloadImportance
```
rido-min: I thought we agree to do implement enums with String Literals, is there any reason to use an enum and the associated converter?

PR #366, opened 2026-03-10
Set Adaptive Cards version number to 1.5

corinagum:
It appears that mobile has a verification step on the Adaptive Cards version being set, despite desktop not having that check. This causes cards not to render if the card does not have a version. Setting a versions fixes rendering on mobile.


heyitsaamir (approved):

corinagum:
Closing for now, I need to make changes internally, properly.

PR #392, opened 2026-03-26
Activity specific builders for outgoing activities

MehakBindra:
What changed: Introduced type-specific activity builders alongside the existing TeamsActivityBuilder.                                        
  New types                                                                                                                                    
  - TeamsActivityBuilder<TActivity, TBuilder> — abstract generic base that all Teams builders inherit from. Holds shared methods: WithChannelData, WithEntities, AddEntity, WithAttachments, AddAttachment, AddAdaptiveCardAttachment, and the Teams-specific overrides for SetConversation/SetFrom/SetRecipient.
  - MessageActivityBuilder — concrete builder for MessageActivity with typed setters for WithText, WithSuggestedActions, and AddMention (reads/writes activity.Text directly).
  - StreamingActivityBuilder — concrete builder for StreamingActivity with typed WithText.
  - TeamsActivityBuilder — now just a thin concrete class with Build() only, backwards-compatible for non-message use cases.

---

  Why
- Previously TeamsActivityBuilder built all activity types but returned TeamsActivity. So you could do WithText("...") but not access activity.Text after building. This is a problem. Even if we internally cast, the user cannot access activity.Text without casting again
  With dedicated builders, MessageActivity.CreateBuilder().WithText(...).Build() returns a MessageActivity — activity.Text just works. Same for SuggestedActions, TextFormat, etc. 
- Since we have limited outgoing activities, this is a clean approach. Developers should directly be accessing Message Activity, and Teams Activity is more of an internal class to coordinate shared properties across outgoing/incoming activities and a bridge to core activity. 

---

  Ripple changes

  - CoreActivityBuilder.WithConversationReference parameter widened from TActivity to CoreActivity so any builder can accept any activity as a conversation reference source.
  - Context.cs and all samples updated to use MessageActivity.CreateBuilder() where they were building message replies.
  - Test suite split into TeamsActivityBuilderTests, MessageActivityBuilderTests, StreamingActivityBuilderTests to match the new structure.

### DIFF
--- a/core/samples/AllFeatures/Program.cs
+++ b/core/samples/AllFeatures/Program.cs
@@ -14,8 +14,7 @@ teamsApp.OnMessage = async (messageArgs, context, cancellationToken) =>
 
     await context.SendTypingActivityAsync(cancellationToken);
 
-    TeamsActivity reply = TeamsActivity.CreateBuilder()
-        .WithType(TeamsActivityType.Message)
+    MessageActivity reply = MessageActivity.CreateBuilder()
         .WithConversationReference(context.Activity)
         .WithText(replyText)
         .Build();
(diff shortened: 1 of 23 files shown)


singhk97 (commented):
Looks good to me

MehakBindra:
Closing in favor of https://github.com/microsoft/teams.net/pull/602 that introduces dedicated outbound activity types

PR #419, opened 2026-04-09
Fixes null reference exception bug when APX returns a 202 response w/o body when sending an activity

singhk97:
**Null-safe Handling of Send Activity Responses:**

* Changed the return type of `SendAsync` in `ActivitiesApi.cs` to `Task<SendActivityResponse?>`, allowing for null responses when the service returns no body.
* Updated `TeamsStreamingWriter.cs` to handle potentially null `SendActivityResponse` objects in both `SendInformativeUpdateAsync` and `AppendResponseAsync`, ensuring `_streamId` is only set if a response is present. [[1]](diffhunk://#diff-19137d244f3805ddd4f0cb3a2f42577ca66bc85f7c8eabb4ab4b38a8e95f363eL75-R76) [[2]](diffhunk://#diff-19137d244f3805ddd4f0cb3a2f42577ca66bc85f7c8eabb4ab4b38a8e95f363eL100-R101)
* Modified `CompatConversations.cs` to handle null `SendActivityResponse` objects when replying to activities, defaulting the resource response ID to an empty string if necessary.


core/src/Microsoft.Teams.Bot.Compat/CompatConversations.cs
```diff
@@ -317,11 +317,11 @@ public async Task<HttpOperationResponse<ResourceResponse>> SendToConversationWit
             // Ensure conversation ID is set
             coreActivity.Conversation ??= new Microsoft.Teams.Bot.Core.Schema.Conversation { Id = conversationId };
 
-            SendActivityResponse response = await _client.SendActivityAsync(coreActivity, convertedHeaders, cancellationToken).ConfigureAwait(false);
+            SendActivityResponse? response = await _client.SendActivityAsync(coreActivity, convertedHeaders, cancellationToken).ConfigureAwait(false);
 
             ResourceResponse resourceResponse = new()
             {
-                Id = response.Id
+                Id = response?.Id ?? string.Empty
```
singhk97, line 324: `string.Empty` to not break api contract. `ResourceResponse.Id` is a string.

rido-min (approved):

PR #422, opened 2026-04-11
fix(stabilization): all 18 audit fixes from STABILIZATION-PLAN.md

rido-min:
## Summary

Implements every fix from `core/docs/stabilization/STABILIZATION-PLAN.md`. All 23 raw audits (consolidated to 18 work items) are addressed with one commit per issue and accompanying unit tests.

### Phase 1 — Critical (production risk)

- **A-012** `fix(A-012)`: eliminate race condition on `OnActivity` in `CompatAdapter` — `ProcessAsync` was overwriting a shared field per-request, corrupting handlers under concurrency. Added a per-request overload to `BotApplication.ProcessAsync`.
- **A-001** `fix(A-001)`: replace blocking `.GetAwaiter().GetResult()` in JWT key resolver — introduced `JwksKeyCache` with background pre-warming so the synchronous `IssuerSigningKeyResolver` serves from a volatile cache without ever blocking a thread-pool thread on the hot path.

### Phase 2 — High (silent data loss / crash)

- **AUTH-01** `fix(AUTH-01)`: `Guid.Parse` → `Guid.TryParse` + fallback; `new JwtSecurityToken` → `JsonWebToken` wrapped in try/catch.
- **SER-01** `fix(SER-01)`: fill 7 missing entries in `ActivitySerializerMap` to match `ActivityDeserializerMap`; register all types in `TeamsActivityJsonContext`.
- **CAST-01** `fix(CAST-01)`: null-safe as-cast fallback in `TeamsActivity` shadowed property getters (From/Recipient/Conversation/ChannelData); also fixed `TeamsConversationAccount.FromConversationAccount` sharing the Properties reference.
- **A-002** `fix(A-002)`: replace fragile `(T)(object)` double-cast in `BotHttpClient` string deserialization path with a typed `DeserializeAsString<T>` helper.
- **DI-01** `fix(DI-01)`: remove two `BuildServiceProvider()` calls during DI registration (`BotConfig.Resolve`, `JwtExtensions.ResolveBotConfig`); fail-fast with `InvalidOperationException` when `IConfiguration` is absent.

### Phase 3 — Medium (correctness defects)

- **A-021** `fix(A-021)`: replace Newtonsoft→STJ two-leg round-trip with single STJ serializer in `SendMeetingNotificationAsync`.
- **A-016** `fix(A-016)`: replace 5 direct `(Activity)` casts with `as Activity ?? throw ArgumentException` in `CompatTeamsInfo`; guard fires before `GetTeamsApiClient` for early, descriptive errors.
- **COPY-01** `fix(COPY-01)`: deep-copy all mutable reference fields in `CoreActivity` copy constructor (via STJ round-trip); fix `CitationEntity` to clone each `CitationClaim`; fix `TeamsConversationAccount.FromConversationAccount` copying Properties by reference.
- **A-014** `fix(A-014)`: seal `TeamsAttachmentBuilder` after `Build()` with a `_built` flag; second call throws `InvalidOperationException`.
- **A-008** `fix(A-008)`: preserve unknown entity types as base `Entity` instead of `null`; fix `ToJsonArray` to handle `JsonElement` extension-data values.
- **A-007** `fix(A-007)`: add `ExtractString(object?)` helper in `CompatActivity` that uses `je.GetString()` for `JsonElement` values instead of `.ToString()` (which includes JSON quotes).
- **A-009** `fix(A-009)`: replace single fragile message-string match with `IsStreamCancelled` helper that checks HTTP status 499, 408, and text as a fallback.

### Phase 4 — Low (hygiene)

- **A-010** `fix(A-010)`: replace `_accumulated += chunk` (O(n²)) with `StringBuilder.Append` in `TeamsStreamingWriter`.
- **A-020** `fix(A-020)`: add `TurnMiddleware.Freeze()` + `BotApplication.FreezeMiddleware()` to seal the pipeline after startup.
- **A-022** `fix(A-022)`: add `ArgumentException.ThrowIfNullOrWhiteSpace(ServiceUrl)` guard in two `CompatConversations` methods that used `new Uri(ServiceUrl!)`.
- **A-019** `fix(A-019)`: remove misleading `GC.SuppressFinalize` from `CompatConnectorClient.Dispose()`.

## Test plan

- [ ] All 18 commits include targeted unit tests
- [ ] **139** Core unit tests pass (net8 + net10)
- [ ] **155** Apps unit tests pass (net8 + net10)
- [ ] **55** Compat unit tests pass (net10)
- [ ] No regressions in any pre-existing test

### DIFF
--- a/core/docs/stabilization/audit-001-blocking-async-jwt.md
+++ b/core/docs/stabilization/audit-001-blocking-async-jwt.md
@@ -1,0 +1,73 @@
+# Audit Issue 001: Blocking Async Call in JWT Signing-Key Resolver
+
+**Severity:** Critical  
+**File:** `core/src/Microsoft.Teams.Bot.Core/Hosting/JwtExtensions.cs`  
+**Line:** 222  
+**Category:** Async/Await correctness
+
+---
+
+## Problem
+
+Inside the `IssuerSigningKeyResolver` delegate — which is invoked synchronously by the ASP.NET Core JWT middleware during token validation — there is a blocking call:
+
+```csharp
+// JwtExtensions.cs, line 222
+OpenIdConnectConfiguration config = manager.GetConfigurationAsync(CancellationToken.None).GetAwaiter().GetResult();
+```
+
+`IssuerSigningKeyResolver` is a synchronous `Func<>` callback. Because ASP.NET Core uses a thread-pool-based async pipeline, calling `.GetAwaiter().GetResult()` here risks **thread-pool starvation**: the thread that calls `GetResult()` is blocked waiting for the async operation to complete, but that operation may itself need a thread-pool thread to continue, leading to a deadlock or severe latency degradation under load.
+
+Additionally, `CancellationToken.None` is passed — if the HTTP fetch for the OIDC metadata hangs, there is no timeout or cancellation path.
+
+---
+
+## Root Cause
+
+`TokenValidationParameters.IssuerSigningKeyResolver` is defined as:
+
+```csharp
+IssuerSigningKeyResolver = (string token, SecurityToken securityToken, string kid, TokenValidationParameters validationParameters)
+    => IEnumerable<SecurityKey>
+```
+
+This is a **synchronous** delegate type. There is no async equivalent in the `Microsoft.IdentityModel.Tokens` library for this callback, so an async fetch must be run synchronously here. The current approach is a common workaround that is known to be unsafe in production ASP.NET Core applications.
+
+---
+
+## Suggested Fix Plan
+
+### Option A — Cache keys eagerly via background refresh (preferred)
+
+Pre-fetch and cache the signing keys in a `Lazy<Task<IEnumerable<SecurityKey>>>` or a background `Timer`, so the synchronous resolver can return from an already-populated in-memory cache without ever blocking:
+
+1. Replace the `ConcurrentDictionary<string, ConfigurationManager<OpenIdConnectConfiguration>>` with a `ConcurrentDictionary<string, Lazy<IEnumerable<SecurityKey>>>`.
+2. On first access, start the async fetch on a **background task** (fire-and-forget with error handling). Return an empty key set (authentication will fail-fast) until the cache warms up.
+3. Schedule periodic refresh (e.g., every 24 hours) to pick up key rotations.
+4. The resolver reads from the cache synchronously — no blocking calls.
+
+### Option B — Use `ConfigurationManager<T>` with synchronous HTTP (acceptable interim)
+
+The `ConfigurationManager<T>` internally uses `HttpDocumentRetriever`. Replace it with a synchronous HTTP call using `HttpClient.Send()` (not `SendAsync`) so the blocking is explicit and understood, rather than disguised as async:
+
+```csharp
+// Temporary mitigation while a full async cache is implemented.
+// Clearly mark this with a TODO comment and a work item.
+var httpClient = new HttpClient();
+var json = httpClient.GetStringAsync(authority).GetAwaiter().GetResult(); // Known blocking call
+```
+
+This does not fix starvation but makes the intent explicit and avoids nested async context issues.
+
+### Option C — Switch to a custom `ISecurityTokenValidator` (cleanest long-term)
+
+Implement a custom `JwtBearerOptions.SecurityTokenValidators` entry or override `OnTokenValidated` to perform async key resolution before the synchronous validator runs, caching keys into `TokenValidationParameters.IssuerSigningKeys` on each request.
+
+---
+
+## Acceptance Criteria
+
+- No `.GetAwaiter().GetResult()` or `.Result` on a `Task` inside the `IssuerSigningKeyResolver` delegate.
+- Signing-key resolution does not block a thread-pool thread under concurrent load.
+- Key refresh still happens — stale keys do not cause permanent authentication failures.
+- Existing integration tests for JWT validation continue to pass.
--- a/core/docs/stabilization/audit-002-unsafe-generic-cast.md
+++ b/core/docs/stabilization/audit-002-unsafe-generic-cast.md
@@ -1,0 +1,104 @@
+# Audit Issue 002: Unsafe `(T)(object)` Cast in Generic HTTP Response Deserializer
+
+**Severity:** Critical  
+**File:** `core/src/Microsoft.Teams.Bot.Core/Http/BotHttpClient.cs`  
+**Lines:** 216, 220  
+**Category:** Type safety
+
+---
+
+## Problem
+
+In `DeserializeResponseAsync<T>`, there is a special-case branch for `T == string` that uses a double-cast to work around C#'s type system:
+
+```csharp
+// BotHttpClient.cs, lines 211-222
+if (typeof(T) == typeof(string))
+{
+    try
+    {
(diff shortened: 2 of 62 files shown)


rajan-chari:
Looked at fix for:

A-012 fix(A-012): eliminate race condition on OnActivity in CompatAdapter — ProcessAsync was overwriting a shared field per-request, corrupting handlers under concurrency. Added a per-request overload to BotApplication.ProcessAsync.

Fixes the race condition.  Looks good to me.

Not sure how to review this PR since there are a lot of changes.  Split this into at least 3? 
Phase 1, 2, 3 + a few more so review burden can be lowered.

core/docs/stabilization/audit-001-blocking-async-jwt.md
```diff
@@ -0,0 +1,73 @@
...
+## Problem
+
+Inside the `IssuerSigningKeyResolver` delegate — which is invoked synchronously by the ASP.NET Core JWT middleware during token validation — there is a blocking call:
+
+```csharp
+// JwtExtensions.cs, line 222
+OpenIdConnectConfiguration config = manager.GetConfigurationAsync(CancellationToken.None).GetAwaiter().GetResult();
+```
+
+`IssuerSigningKeyResolver` is a synchronous `Func<>` callback. Because ASP.NET Core uses a thread-pool-based async pipeline, calling `.GetAwaiter().GetResult()` here risks **thread-pool starvation**: the thread that calls `GetResult()` is blocked waiting for the async operation to complete, but that operation may itself need a thread-pool thread to continue, leading to a deadlock or severe latency degradation under load.
```
singhk97, line 19: `GetAwaiter().GetResult()` blocks the thread-pool thread and so we should also look at other parts of the codebase to flag this pattern.

core/docs/stabilization/audit-001-blocking-async-jwt.md
```diff
@@ -0,0 +1,73 @@
...
+---
+
+## Suggested Fix Plan
+
+### Option A — Cache keys eagerly via background refresh (preferred)
+
+Pre-fetch and cache the signing keys in a `Lazy<Task<IEnumerable<SecurityKey>>>` or a background `Timer`, so the synchronous resolver can return from an already-populated in-memory cache without ever blocking:
+
+1. Replace the `ConcurrentDictionary<string, ConfigurationManager<OpenIdConnectConfiguration>>` with a `ConcurrentDictionary<string, Lazy<IEnumerable<SecurityKey>>>`.
+2. On first access, start the async fetch on a **background task** (fire-and-forget with error handling). Return an empty key set (authentication will fail-fast) until the cache warms up.
```
singhk97, line 45: We should be able to force a cache warm up instead of failing on the first try.

core/docs/stabilization/audit-001-blocking-async-jwt.md
```diff
@@ -0,0 +1,73 @@
...
+3. Schedule periodic refresh (e.g., every 24 hours) to pick up key rotations.
```
singhk97, line 46: We should sync this with known key rotation time periods

core/docs/stabilization/audit-002-unsafe-generic-cast.md
```diff
@@ -0,0 +1,104 @@
+# Audit Issue 002: Unsafe `(T)(object)` Cast in Generic HTTP Response Deserializer
+
+**Severity:** Critical
```
singhk97, line 3: This doesn't seem critical

core/docs/stabilization/audit-003-asymmetric-serializer-maps.md
```diff
@@ -0,0 +1,109 @@
+# Audit Issue 003: Asymmetric Serializer / Deserializer Maps Causing Silent Data Loss
```
singhk97, line 1: This is a feature not a bug. We wouldn't want to serialize a `MessageUpdate`/`MessageDelete`/`ConversationUpdate`/`InstallationUpdate ` activity

core/docs/stabilization/audit-004-property-shadow-as-casts.md
```diff
@@ -0,0 +1,138 @@
...
+    get => base.ChannelData as TeamsChannelData;
+    set => base.ChannelData = value;
+}
+```
+
+The `as` operator returns `null` if the object is not of the target type — **it never throws**. This creates two categories of silent failure:
+
+1. **Type mismatch at runtime** — If the base property was set to a `ConversationAccount` (not a `TeamsConversationAccount`), `From` returns `null` with no diagnostic. Callers that assume the property is non-null (because it was set) receive `null` and may NullReferenceException later, far from the assignment site.
+
+2. **Polymorphism violation** — Code holding a `CoreActivity` reference gets different property values than code holding a `TeamsActivity` reference to the _same object_. This violates the Liskov Substitution Principle and can cause subtle bugs in generic code that accepts `CoreActivity`.
```
singhk97, line 49: This is problematic b.c. `TeamsActivity` (i assume) is used in general methods that accept a `CoreActivity` parameter

core/docs/stabilization/audit-005-build-service-provider.md
```diff
@@ -0,0 +1,109 @@
+# Audit Issue 005: `BuildServiceProvider()` Called Multiple Times During Startup
```
singhk97, line 1: Good catch

singhk97 (commented):
Reviewed audit 1-6

core/docs/stabilization/audit-001-blocking-async-jwt.md
rido-min: that's what the audit did, seems this is the only place in core (there are more in main)

Can you resolve this comment?

core/docs/stabilization/audit-001-blocking-async-jwt.md
rido-min: I'm not able to make it fail on the first request, note that the suggested implementation is not using Lazy, can you review the actual fix?

core/docs/stabilization/audit-001-blocking-async-jwt.md
rido-min: I'm not sure how often they rotate the keys, and we will need to get into long-haul testing to verify. worst case, an app restart will fix it.. no?

I'm not sure how the keys are refreshed in 2.0/2.1 before this fix, but having a blocking call in a hot code path does not seems right to me.

core/docs/stabilization/audit-002-unsafe-generic-cast.md
rido-min: For a library code, where we accept any T, this might be a problem, specially when T is string.

I dont see any issue with the suggested fix.

core/docs/stabilization/audit-003-asymmetric-serializer-maps.md
rido-min: agree on the intent, but the fix just avoids issues if a customer call toJson from onMessageDelete. I dont see much harm, but @MehakBindra  might have more insights.

core/docs/stabilization/audit-004-property-shadow-as-casts.md
rido-min: The real issue is with the shadow properties, and rebase. I'm thinking to remove channelData from CoreActivity, so we won't have these issues

singhk97 (approved):

PR #433, opened 2026-04-20
Prompt Preview Support

ShanmathiMayuramKrithivasan:
Prompt preview allows a bot to reference the original targeted message in its reply, so Teams can render a collapsible preview of the user's prompt.

When a bot replies to a targeted message (reactive scenarios), the SDK now passes the original message's messageId via a targetedMessageInfo entity in the activity's entities array.

For proactive replies, developers can attach the entity themselves with the messageId of the targeted message.


Libraries/Microsoft.Teams.Apps/Contexts/Context.Send.cs
```diff
@@ -67,6 +68,20 @@ public partial class Context<TActivity> : IContext<TActivity>
 {
     public async Task<T> Send<T>(T activity, CancellationToken cancellationToken = default) where T : IActivity
     {
+        // Auto-populate targetedMessageInfo entity for prompt preview
+        // when the incoming activity was a targeted message (reactive flow).
+        #pragma warning disable ExperimentalTeamsTargeted
+        if (Activity.Recipient?.IsTargeted == true && Activity.Id is not null)
```
corinagum: The guard is inside the generic `Send<T>` where `T : IActivity`, so this fires for every activity type — `TypingActivity`, `EndOfConversationActivity`, etc. — not just messages. APX only consumes `targetedMessageInfo` inside `MessageEncoder` (gated on `EnablePromptPreview`), so non-message sends during a targeted session will carry a stray entity that APX silently ignores.

Not a correctness bug, but wasted payload + log noise. Consider narrowing the guard:
```csharp
if (activity is MessageActivity && Activity.Recipient?.IsTargeted == true && Activity.Id is not null)
```

Libraries/Microsoft.Teams.Api/Entities/TargetedMessageInfoEntity.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Teams.Api.Entities;
+
+public class TargetedMessageInfoEntity : Entity
```
corinagum, line 10: This new entity depends on `Account.IsTargeted`, which is marked `[Experimental("ExperimentalTeamsTargeted")]` (Account.cs:40). Other Targeted-Message surfaces carry the same attribute — `ActivityClient.cs:103,128,151`, `MessageActivity.cs:155`, `Activity.cs:234`. The auto-populate call site in `Context.Send.cs` wraps the usage in `#pragma warning disable ExperimentalTeamsTargeted`, which itself indicates the feature is still experimental.

For consistency with the rest of the Targeted-Message API surface, consider marking the class (and/or `MessageId`) `[Experimental("ExperimentalTeamsTargeted")]` so consumers opt in explicitly until the feature stabilizes.

corinagum (changes requested):
- Is quoted reply and prompt preview stacking the intended UX? What should this look like?
- Has it been tested? How can we test this out?

Libraries/Microsoft.Teams.Apps/Contexts/Context.Send.cs
ShanmathiMayuramKrithivasan: Fixed.

Libraries/Microsoft.Teams.Api/Entities/TargetedMessageInfoEntity.cs
ShanmathiMayuramKrithivasan: Fixed.

Libraries/Microsoft.Teams.Apps/Contexts/Context.Send.cs
```diff
@@ -67,6 +68,20 @@ public partial class Context<TActivity> : IContext<TActivity>
...
+        if (activity is MessageActivity && Activity.Recipient?.IsTargeted == true && Activity.Id is not null)
+        {
+            var hasEntity = activity.Entities?.Any(e => e is TargetedMessageInfoEntity) ?? false;
+            if (!hasEntity)
+            {
+                activity.Entities ??= [];
+                activity.Entities.Add(new TargetedMessageInfoEntity { MessageId = Activity.Id });
+            }
+        }
+        #pragma warning restore ExperimentalTeamsTargeted
```
corinagum, line 77: Also scan `activity.Entities` for any entity whose wire type equals `"quotedReply"` targeting `Activity.Id`, and scan `activity.Text` for a matching `<quoted messageId="<tm-id>"/>` placeholder. If either is found, strip both and log. Per §6.1 of the prompt preview design, PP owns the preview surface for targeted message flows and `qtdMsgs` must not be set. Matching on the string `"quotedReply"` avoids taking a dependency on the quoted replies PR, which has not landed on main.

Libraries/Microsoft.Teams.Api/Entities/TargetedMessageInfoEntity.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Teams.Api.Entities;
+
```
corinagum, line 8: A parallel `AddTargetedMessageInfo(string messageId)` on `Activity` would match `AddFeedback` / `AddCitation` and the TS `addTargetedMessageInfo` already in this PR. Gives a single home for the §1 P0 "one PP per message" dedup plus the usage XML doc.

ShanmathiMayuramKrithivasan: Updated.

Libraries/Microsoft.Teams.Apps/Contexts/Context.Send.cs
ShanmathiMayuramKrithivasan: Fixed, but from the other direction - Reply skips ToQuoteReply when the incoming activity is targeted, so the blockquote is never generated. Same outcome. Since quoted reply is still not fully rolled out, going with this approach for the fix.

corinagum: Not a fan of this approach. I intentionally asked for the check to be for quotedReply here to avoid collision with the quoted replies sdk changes, and we don't know which order they will be merged in.

ShanmathiMayuramKrithivasan:
> * Is quoted reply and prompt preview stacking the intended UX? What should this look like?
> * Has it been tested? How can we test this out?

Yes that is the intended behavior. If both QR and PP are sent, they are both stacked in the UX. 
I have shared testing steps in Teams. Also sample app in this PR is updated to validate the changes.

Libraries/Microsoft.Teams.Apps/Contexts/Context.Send.cs
ShanmathiMayuramKrithivasan: Updated - AddTargetedMessageInfo now guards against QR collision by stripping any "quotedReply" entities and matching <quoted .../> text placeholder before adding PP. No dependency on the quoted replies PR. Reverted the Reply skip.

Libraries/Microsoft.Teams.Apps/Contexts/Context.Send.cs
```diff
@@ -67,6 +67,15 @@ public partial class Context<TActivity> : IContext<TActivity>
 {
     public async Task<T> Send<T>(T activity, CancellationToken cancellationToken = default) where T : IActivity
     {
+        // Auto-populate targetedMessageInfo entity for prompt preview
+        // when the incoming activity was a targeted message (reactive flow).
+        #pragma warning disable ExperimentalTeamsTargeted
+        if (activity is MessageActivity messageActivity && Activity.Recipient?.IsTargeted == true && Activity.Id is not null)
+        {
+            messageActivity.AddTargetedMessageInfo(Activity.Id);
```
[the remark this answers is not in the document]
ShanmathiMayuramKrithivasan: IsTargeted is only populated by the Teams service on incoming message activities from the targeted messages flow - invoke/reaction activities won't have this flag set. The existing guard (Activity.Recipient?.IsTargeted == true) is sufficient to scope this to the correct scenario.

Libraries/Microsoft.Teams.Api/Activities/Activity.cs
```diff
@@ -427,6 +427,42 @@ public virtual Activity AddFeedback(FeedbackType mode)
...
+                {
+                    if (Entities[i].Type == "quotedReply") Entities.RemoveAt(i);
+                }
+            }
+
+            if (this is MessageActivity message && message.Text is not null)
+            {
+                message.Text = message.Text.Replace($"<quoted messageId=\"{messageId}\"/>", string.Empty).Trim();
+            }
+
```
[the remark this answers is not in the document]
ShanmathiMayuramKrithivasan: Addressed

corinagum (approved):

PR #442, opened 2026-04-22
Update next/core with changes from main

rido-min:
This is a RI to make sure next/core has no conflicts with main


rajan-chari (approved):
Ack.  This is main -> core merge.

PR #460, opened 2026-04-29
Core/preview stabilization

rido-min:


core/src/Microsoft.Teams.Apps/Schema/TeamsActivityJsonContext.cs
```diff
@@ -19,6 +20,13 @@ namespace Microsoft.Teams.Apps.Schema;
 [JsonSerializable(typeof(TeamsActivity))]
 [JsonSerializable(typeof(MessageActivity))]
 [JsonSerializable(typeof(StreamingActivity))]
+[JsonSerializable(typeof(MessageReactionActivity))]
```
MehakBindra, line 23: These were intentionally omitted since these activities are never sent and never serialized. I think we should not add these.

core/src/Microsoft.Teams.Apps/Schema/TeamsActivityType.cs
```diff
@@ -100,5 +100,11 @@ public static class TeamsActivityType
     {
         [typeof(MessageActivity)] = a => a.ToJson(TeamsActivityJsonContext.Default.MessageActivity),
         [typeof(StreamingActivity)] = a => a.ToJson(TeamsActivityJsonContext.Default.StreamingActivity),
+        [typeof(MessageReactionActivity)] = a => a.ToJson(TeamsActivityJsonContext.Default.MessageReactionActivity),
```
MehakBindra, line 103: Let's not add code that will never be used - these activities are never serialized

core/src/Microsoft.Teams.Core/Schema/CoreActivity.cs
```diff
@@ -30,6 +40,28 @@ public class ExtendedPropertiesDictionary : Dictionary<string, object?>
 
         return default;
     }
+
+    /// <summary>
+    /// Gets and deserializes a value from the dictionary without removing it.
+    /// Handles <see cref="System.Text.Json.JsonElement"/> values that result from deserialization.
+    /// </summary>
+    public T? Get<T>(string key)
```
MehakBindra, line 48: This is duplicating "Extract" method above, need to consolidate

core/src/Microsoft.Teams.Apps/Schema/Entities/Entity.cs
```diff
@@ -59,18 +59,14 @@ public class EntityList : List<Entity>
                 && typeValue.GetValue<string>() is string typeString)
             {
 
-                // TODO: Should be able to support unknown types (PA uses BotMessageMetadata).
-                // TODO: Investigate if there is any way for Parent to avoid
```
MehakBindra, line 63: don't delete comment that is still relevant

core/src/Microsoft.Teams.Core/Hosting/BotConfig.cs
```diff
@@ -142,10 +142,22 @@ public static BotConfig Resolve(IServiceCollection services, string sectionName
...
         ArgumentNullException.ThrowIfNull(services);
 
-        // Extract IConfiguration from service collection
+        // Extract IConfiguration from service collection — prefer the instance if available,
+        // otherwise resolve via the factory registered in the descriptor.
         ServiceDescriptor? configDescriptor = services.FirstOrDefault(d => d.ServiceType == typeof(IConfiguration));
-        IConfiguration configuration = configDescriptor?.ImplementationInstance as IConfiguration
-            ?? services.BuildServiceProvider().GetRequiredService<IConfiguration>();
+        IConfiguration? configuration = configDescriptor?.ImplementationInstance as IConfiguration;
+        if (configuration is null && configDescriptor?.ImplementationFactory is not null)
```
MehakBindra, line 149: This is still using BuildServiceProvider anti-pattern? Tried removing in a previous commit - https://github.com/microsoft/teams.net/pull/460/changes/0d75cb306fa5a66e2927c81b5f99771436acbda5 . Maybe needs a human decision on what is best

core/src/Microsoft.Teams.Core/Schema/CoreActivity.cs
rido-min: They share the same lookup + JsonElement deserialization logic, but the post-deserialization behavior is intentionally different:
  - Extract<T> removes the entry (so properties promoted to typed fields don't also live in the dictionary)
  - Get<T> keeps the entry and replaces the raw JsonElement with the deserialized object (so the next Get<T> call skips deserialization)

  These aren't duplicates — they're complementary. I'd recommend keeping them separate. If you'd like to reduce the code duplication in their shared lookup/deserialization logic, we could extract a private
  helper, but the public API surface should stay as two distinct methods.

MehakBindra: Thanks for explaining !

core/src/Microsoft.Teams.Apps/Schema/TeamsActivityJsonContext.cs
singhk97: agreed

core/src/Microsoft.Teams.Apps/Schema/TeamsActivityType.cs
singhk97: agreed

core/src/Microsoft.Teams.Apps.BotBuilder/CompatTeamsInfo.cs
```diff
@@ -356,7 +356,7 @@ private static AgenticIdentity GetIdentity(ITurnContext turnContext)
         AgenticIdentity agenticIdentity = GetIdentity(turnContext);
 
         string url = $"{serviceUrl.ToString().TrimEnd('/')}/v1/meetings/{Uri.EscapeDataString(meetingId)}/notification";
-        string body = JsonSerializer.Serialize(notification);
+        string body = JsonConvert.SerializeObject(notification);
```
singhk97, line 359: This is a massive bug fix. Good find!

core/src/Microsoft.Teams.Apps.BotBuilder/CompatTeamsInfo.cs
```diff
@@ -546,7 +546,9 @@ public static async Task<string> SendMessageToAllUsersInTeamAsync(
         ConversationClient client = GetConversationClient(turnContext);
         Uri serviceUrl = new(GetServiceUrl(turnContext));
         AgenticIdentity agenticIdentity = GetIdentity(turnContext);
-        CoreActivity coreActivity = ((Activity)activity).FromCompatActivity();
+        if (activity is not Activity teamActivity)
+            throw new ArgumentException("Expected a Bot Framework Activity instance.", nameof(activity));
```
singhk97, line 550: it solves a problem that doesn't exist lol

core/src/Microsoft.Teams.Apps/Schema/Entities/Entity.cs
singhk97: looks like PA TODO is fixed, not sure about the other one

core/src/Microsoft.Teams.Apps.BotBuilder/CompatActivity.cs
```diff
@@ -63,37 +72,37 @@ public static Microsoft.Bot.Schema.ChannelAccount ToCompatChannelAccount(this Mi
 
         if (account.Properties.TryGetValue("aadObjectId", out object? aadObjectId))
         {
-            channelAccount.AadObjectId = aadObjectId?.ToString();
+            channelAccount.AadObjectId = GetStringValue(aadObjectId);
```
singhk97, line 75: supporting flexible aadObjectId is a good idea imo

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -113,7 +114,9 @@ public async Task AppendResponseAsync(string chunk, CancellationToken cancellati
             _streamId ??= response?.Id;
             _lastChunkSent = DateTime.UtcNow;
         }
-        catch (HttpRequestException ex) when (ex.Message.Contains("Content stream was cancelled by user", StringComparison.OrdinalIgnoreCase))
+        catch (HttpRequestException ex) when (
+            ex.StatusCode is HttpStatusCode.Gone or HttpStatusCode.NoContent
```
singhk97, line 118: did we manually verify this? i believe the message string could be tied to what is returned by APX

rido-min:
superseed by #462

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
rido-min: no I did not verify, I thought this was your fix to solve the 201s with empty content.. no?

As this PR is closed we should add a note somewhere to review

MehakBindra: || ex.Message.Contains("Content stream was cancelled", StringComparison.OrdinalIgnoreCase)) is still there, so no loss

PR #489, opened 2026-05-07
Add proactive threading helpers to core (port of #425)

corinagum:
## Summary
- Adds `Microsoft.Teams.Core.Schema.ConversationExtensions` with a `ThreadId(this Conversation)` extension method and a `ToThreadedConversationId(conversationId, messageId)` static validator (rejects non-numeric / zero `messageId`, strips any existing `;messageid=` suffix). Ported from `Libraries/Microsoft.Teams.Api/Conversation.cs`, where they live as instance/static members on `Conversation`; on core they live as a separate extension class so `Conversation` stays a pure POCO.
- Replaces the inline `$"{conversationId};messageid={messageId}"` interpolation in proactive `TeamsBotApplication.ReplyAsync` with the validating helper.
- Adds `core/samples/Threading/` modeled on `Samples/Samples.Threading/` to exercise reactive (`context.ReplyAsync` / `context.SendActivityAsync`) and proactive (`teamsApp.ReplyAsync` / `ConversationExtensions.ToThreadedConversationId` + `teamsApp.SendAsync`) paths.
- Includes #493 (Rido), stacked on this PR: adds `Async`-suffixed methods to `Context` (`SendAsync`, `ReplyAsync`, `TypingAsync`, `SignInAsync`, `SignOutAsync`) and `TeamsBotApplication` (`SendAsync`, `ReplyAsync`) with the original names kept as backward-compat redirects; proactive `Send`/`Reply` accept an optional `AgenticIdentity?` for user-delegated token acquisition.

## Scope notes
- **`Context.ReplyAsync`** inherits the post-QR (PR #477) behavior on `main`: delegates to `Quote(Activity.Id, ...)` when the inbound activity has an id, producing a `<blockquote>`-rendered quoted reply. This PR doesn't change that delegation.
- **Personal / 1:1 chats**: `test proactive` and `test manual` currently fail at the service with `BadArgument: Failed to decrypt conversation id` until the relevant ECS rollout completes. Channel scopes work today. `core/samples/Threading/README.md` documents this caveat.

## Test plan
- [x] Unit: 12 `ConversationTests` (validator + `ThreadId` extension method)
- [x] Unit: 3 `TeamsBotApplicationTests` (validator propagates from proactive `ReplyAsync`)
- [x] Local CI: core-ci pipeline (325 tests on net8.0 + net10.0) + root build-test-lint / ci.yaml (813 tests + `dotnet pack`); 0 errors
[...]

PR #493, opened 2026-05-08
Add Async suffixes and agentic identity for proactive messaging

rido-min:
## Summary
- Add `Async`-suffixed methods to `Context` (`SendAsync`, `ReplyAsync`, `TypingAsync`, `SignInAsync`, `SignOutAsync`) and `TeamsBotApplication` (`SendAsync`, `ReplyAsync`), keeping original names as redirects for backward compatibility.
- Proactive `Send`/`Reply` on `TeamsBotApplication` now accept an `AgenticIdentity?` parameter (matching the pattern in `ConversationClient`), so user-delegated token acquisition works when running with agentic identities.
- Threading sample updated to extract `AgenticIdentity` from the inbound activity's `Recipient` and pass it to proactive calls.

Stacked on #489.

## Test plan
- [ ] Verify `test proactive` and `test manual` commands work with agentic identities
- [ ] Verify reactive `test reply` and `test send` still work (no change in behavior)
- [ ] Verify existing callers using `Send`/`Reply` without `agenticIdentity` still compile and work

### DIFF
--- a/core/samples/Threading/Program.cs
+++ b/core/samples/Threading/Program.cs
@@ -17,6 +17,10 @@ teamsApp.OnMessage(async (context, cancellationToken) =>
     string conversationId = context.Activity.Conversation!.Id;
     string messageId = context.Activity.Id!;
 
+    // Store the agentic identity from the inbound activity's Recipient (the bot).
+    // Required for proactive messaging when running with agentic identities.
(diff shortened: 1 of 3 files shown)


corinagum (approved):

corinagum:
Thank you @rido-min!

PR #484, opened 2026-05-06
Standardize configuration

MehakBindra:
This pull request refactors the Teams Bot Core authentication and client registration pipeline to standardize configuration handling, simplify service registration, and remove legacy configuration formats. The main focus is on consolidating configuration to use a single Microsoft.Identity.Web-compatible section (typically `AzureAd`), cleaning up old code paths, and improving the extensibility and maintainability of authentication and client setup.

Key changes include:

**Configuration Standardization and Documentation:**
- Updated the architecture documentation to clarify that only Microsoft.Identity.Web-style configuration (e.g., `AzureAd` section) is supported, and removed references to legacy Bot Framework and Core environment variable formats. The configuration precedence and example have been revised accordingly.
- Removed legacy configuration blocks from the TeamsBot sample's launch settings, aligning sample usage with the new configuration standard.

**Service Registration and Pipeline Simplification:**
- Refactored `AddBotApplication`, `AddConversationClient`, and `AddUserTokenClient` extension methods to use a unified configuration section default (`BotConfig.DefaultSectionName`), and to always set up MSAL services and client registration in a consistent way.
- Introduced `EnsureMsalServices` to centralize MSAL and token acquisition pipeline registration, ensuring idempotency and correct configuration binding.

**Code Cleanup and Removal of Legacy Paths:**
- Removed the `BotClientOptions` class and all related code, as options are now passed directly from `BotConfig` and configuration sections.
- Removed all legacy `FromBFConfig` and `FromCoreConfig` methods and associated code from `BotConfig`, so only Microsoft.Identity.Web configuration is supported.

**Handler and Option Refactoring:**
- Simplified the `BotAuthenticationHandler` constructor and logic to remove support for managed identity options, since these are now handled through standard configuration.

PR #495, opened 2026-05-08
Hardcode auth scopes in BotAuthenticationHandler

rido-min:
> [!NOTE]
> Stacked on #484

## Summary
- Remove `scope` parameter from `BotAuthenticationHandler` constructor
- Agentic identities always use `https://botapi.skype.com/.default`
- Non-agentic (app-only) always uses `https://api.botframework.com/.default`

## Test plan
- [ ] Verify app-only token acquisition uses the correct scope
- [ ] Verify agentic token acquisition uses the correct scope


singhk97 (approved):
lgtm

PR #505, opened 2026-05-13
sync main with patches in releases/core

singhk97:
applied some patches to releases/core to fix the released version. syncing main now.


core/src/Microsoft.Teams.Apps/version.json
```diff
@@ -1,6 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/main/src/NerdBank.GitVersioning/version.schema.json",
   "version": "2.1.0-preview.{height}",
+  "versionHeightOffset": -3,
```
corinagum, line 4: Why are these needed? If we're doing squash merges to a release branch, no offset needed right?

singhk97: the release branch already had a few commits on top of the version.json when i had first created it. The offset was to basically undo that. This won't have to be changed again though just for first time initialization.

.azdo/publish.yaml
```diff
@@ -114,7 +114,7 @@ extends:
                 packageParentPath: '$(Build.ArtifactStagingDirectory)'
                 publishVstsFeed: '$(System.TeamProject)/TeamsSDKPreviews'
                 nuGetFeedType: internal
-                allowPackageConflicts: false  # Disallow duplicate versions; bump the version to re-publish
+                allowPackageConflicts: true  # Treat 409 (version already in feed, including deleted versions) as success
```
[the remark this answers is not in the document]
singhk97: this is to ensure that the entire operation doesn't fail just because there is an existing version that already exists in the internal feed.

corinagum (approved):

PR #545, opened 2026-06-02
Validate serviceurl claim matches activity ServiceUrl

rido-min:
Add validation to ensure the "serviceurl" claim in user claims matches the activity's ServiceUrl (case-insensitive) in BotApplication. Throw InvalidDataException on mismatch. Add unit tests covering matching, case-insensitivity, mismatches, and absence of the claim.

### DIFF
--- a/core/src/Microsoft.Teams.Core/BotApplication.cs
+++ b/core/src/Microsoft.Teams.Core/BotApplication.cs
@@ -206,6 +206,22 @@ public class BotApplication
             throw new InvalidDataException($"ServiceUrl in activity ({activity.ServiceUrl}) does not match serviceUrl claim ({serviceUrlFromClaims}).");
         };
 
+        KeyValuePair<string, object?> activityTypeTag = new(Telemetry.Tags.ActivityType, activity.Type);
+        Telemetry.ActivitiesReceived.Add(1, activityTypeTag);
+
+        using Activity? span = Telemetry.Source.StartActivity(Telemetry.Spans.Turn, ActivityKind.Internal);
+        if (span is not null)
+        {
+            span.SetTag(Telemetry.Tags.ActivityType, activity.Type);
+            span.SetTag(Telemetry.Tags.ActivityId, activity.Id);
+            span.SetTag(Telemetry.Tags.ConversationId, activity.Conversation?.Id);
+            span.SetTag(Telemetry.Tags.ChannelId, activity.ChannelId);
+            span.SetTag(Telemetry.Tags.BotId, AppId);
+            span.SetTag(Telemetry.Tags.ServiceUrl, activity.ServiceUrl?.ToString());
+        }
+
+        long startTimestamp = Stopwatch.GetTimestamp();
+
         // TODO: Replace with structured scope data, ensure it works with OpenTelemetry and other logging providers
         using (_logger.BeginActivityScope(activity.Type, activity.Id, activity.ServiceUrl, correlationVector))
         {
--- a/core/src/Microsoft.Teams.Core/BotApplication.cs
+++ b/core/src/Microsoft.Teams.Core/BotApplication.cs
@@ -201,7 +201,7 @@ public class BotApplication
         }
 
         string serviceUrlFromClaims = httpContext.User.Claims.FirstOrDefault(c => c.Type == "serviceurl")?.Value ?? string.Empty;
-        if (!string.IsNullOrEmpty(serviceUrlFromClaims) && !serviceUrlFromClaims.Equals(activity.ServiceUrl?.ToString(), StringComparison.OrdinalIgnoreCase))
+        if (!string.IsNullOrEmpty(serviceUrlFromClaims) && !serviceUrlFromClaims.Equals(activity.ServiceUrl?.ToString(), StringComparison.Ordinal))
         {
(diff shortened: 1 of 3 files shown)


singhk97 (approved):

heyitsaamir (commented):
It'd be good to double check if this logic works in various settings:
1. 1:1
2. Group Chats
3. Channels
4. Threads

core/src/Microsoft.Teams.Core/BotApplication.cs
```diff
@@ -200,6 +200,12 @@ public virtual async Task ProcessAsync(HttpContext httpContext, CancellationToke
             _logger.ReceivedActivityJson(activity.ToJson());
         }
 
+        string serviceUrlFromClaims = httpContext.User.Claims.FirstOrDefault(c => c.Type == "serviceurl")?.Value ?? string.Empty;
+        if (!string.IsNullOrEmpty(serviceUrlFromClaims) && !serviceUrlFromClaims.Equals(activity.ServiceUrl?.ToString(), StringComparison.OrdinalIgnoreCase))
```
heyitsaamir: Instead of matching exactly, should we strip whitespaces/slashes atleast? We should also double check if the url is expected to be _exactly_ the same or just the hostname needs to match.

rido-min: why not exactly? any string manipulation (removing whitespaces, slashes) might result in another attack vector. Both values are set by the same service and those should match exactly (thinking to force casing)

heyitsaamir: Sounds good

heyitsaamir (approved):

core/src/Microsoft.Teams.Core/BotApplication.cs
```diff
@@ -200,6 +200,12 @@ public virtual async Task ProcessAsync(HttpContext httpContext, CancellationToke
...
+        {
+            throw new InvalidDataException($"ServiceUrl in activity ({activity.ServiceUrl}) does not match serviceUrl claim ({serviceUrlFromClaims}).");
```
corinagum: I think it's not advised to repeat potentially unsafe urls in logs/errors.

core/src/Microsoft.Teams.Core/BotApplication.cs
```diff
@@ -200,6 +200,12 @@ public virtual async Task ProcessAsync(HttpContext httpContext, CancellationToke
...
+        };
```
corinagum:
```suggestion
        }
```
nit

core/src/Microsoft.Teams.Core/BotApplication.cs
corinagum: `Uri.ToString()` would normalize to have a trailing slash `/`. Do we want to normalize the same way?

core/src/Microsoft.Teams.Core/BotApplication.cs
rido-min: what's your suggestion? we could remove the potentially unsafe url from the exception and add a log.Debug.

core/src/Microsoft.Teams.Core/BotApplication.cs
```diff
@@ -200,22 +200,13 @@ public virtual async Task ProcessAsync(HttpContext httpContext, CancellationToke
...
-            span.SetTag(Telemetry.Tags.ActivityId, activity.Id);
-            span.SetTag(Telemetry.Tags.ConversationId, activity.Conversation?.Id);
-            span.SetTag(Telemetry.Tags.ChannelId, activity.ChannelId);
-            span.SetTag(Telemetry.Tags.BotId, AppId);
-            span.SetTag(Telemetry.Tags.ServiceUrl, activity.ServiceUrl?.ToString());
-        }
-
-        long startTimestamp = Stopwatch.GetTimestamp();
+            throw new InvalidDataException($"ServiceUrl in activity ({activity.ServiceUrl}) does not match serviceUrl claim ({serviceUrlFromClaims}).");
+        };
```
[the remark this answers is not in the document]
rido-min: outdated

core/src/Microsoft.Teams.Core/BotApplication.cs
rido-min: changed in latest commit

singhk97 (approved):

core/src/Microsoft.Teams.Core/BotApplication.cs
rido-min: mmhhh... I dont see Uri.ToString() to add or remove any trailing slash.

do you have a code repro?

rido-min:
> It'd be good to double check if this logic works in various settings:
> 
> 1. 1:1
> 2. Group Chats
> 3. Channels
> 4. Threads

how so? this is only about claims and serviceUrls, it should not affect any other fields in the payload (and yes, tested without any issue)

core/src/Microsoft.Teams.Core/BotApplication.cs
corinagum: I actually meant 'we should use Uri.ToString()' but after reviewing it, it only appends `/` in empty path case. Please disregard.

corinagum (approved):

PR #548, opened 2026-06-03
Enable MSAL agentic token caching in BotAuthenticationHandler

rido-min:
workaround for https://github.com/AzureAD/microsoft-identity-web/issues/3840

Previously, agentic token requests always triggered network calls to Entra due to passing a null ClaimsPrincipal, bypassing MSAL's silent token cache. This change introduces a per-agentic-identity ClaimsPrincipal cache, enabling silent token acquisition and reducing redundant network latency. Also adds a markdown investigation documenting the root cause, impact, and solution rationale.

### DIFF
--- a/core/docs/agentic-token-caching-investigation.md
+++ b/core/docs/agentic-token-caching-investigation.md
@@ -1,0 +1,193 @@
+# Agentic Token Caching Investigation
+
+> **Date**: 2026-06-03
+> **Context**: AppInsights trace `d38cdd6867ffdb64` from A365TeamsAgent shows 4 HTTP calls to `login.microsoftonline.com` per single incoming message, adding ~1s of latency.
+
+## Problem Statement
+
+When processing a single agentic message, `BotAuthenticationHandler` acquires tokens for `https://botapi.skype.com/.default` via the agentic (User FIC / ROPC) flow. MSAL goes to the Entra network endpoint **on every outbound HTTP request**, even though the token is still valid. The in-memory cache is never consulted for the final token exchange.
+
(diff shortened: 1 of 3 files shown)


singhk97 (approved):

rido-min:
Addressed remaining open review comments in 2a0d02ea:

- ** 6, 7 (doc — Option C bullets out of sync with implementation):** Rewrote the *This is the most promising approach* and *Risks* sections of Option C to reflect the shipped design (MemoryCache + per-key semaphore + post-eviction callback), and added a new **Implementation** section after **Recommendation** that documents the IHttpClientFactory `HandlerLifetime` (~2 min) caveat and the verified test coverage.
- ** 8 (semaphores removed during eviction are never disposed):** This concern is rolled into #9 below. The shipped post-eviction callback intentionally does **not** dispose the semaphore — an in-flight request may still hold it and must be able to call `Release()` without `ObjectDisposedException`. GC reclaims the semaphore once all references are released (it allocates no unmanaged handles unless `AvailableWaitHandle` is touched, which this code path never does).
- **  9 (`_agenticLocks` unbounded):** Registered a `PostEvictionCallback` on every cache entry that calls `_agenticLocks.TryRemove(key, …)`. `_agenticPrincipalCache` and `_agenticLocks` now shrink together, restoring the cache bound. Verified by the renamed test `AgenticRequest_CacheEviction_RemovesLockEntry_AndSubsequentCallSucceeds` and the existing `AgenticRequest_ConcurrentCallsDuringCacheEviction_DoNotThrow` test (which now also exercises the no-disposal-on-eviction guarantee).

All 152 tests in `Microsoft.Teams.Core.UnitTests` pass on net8.0 and net10.0.

PR #583, opened 2026-06-26
Add optional TenantId to AgenticIdentity

heyitsaamir:
## Summary

Adds an optional tenant id to `AgenticIdentity`, mirroring [`tenant_id` in teams.py](https://github.com/microsoft/teams.py/blob/b65a9976d5a6f15a5f6c4130ce9f0665824e6a40/packages/api/src/microsoft_teams/api/models/agentic_identity.py#L15).

The tenant arrives on the wire as `tenantId` on the inbound activity's account (alongside `agenticAppId` / `agenticUserId` / `agenticAppBlueprintId`), so the change wires it through end-to-end.

## Changes

- **`AgenticIdentity`** — new optional `TenantId` property; `FromAccount` now maps it.
- **`ChannelAccount`** — added typed `tenantId` JSON property (the wire source); `GetAgenticIdentity()` now populates `TenantId`. A `tenantId` alone still does not fabricate an agentic identity.
- **`TeamsChannelAccount`** — removed its duplicate `tenantId`, now inherited from the base `ChannelAccount` (same JSON name, no behavior change). Updated the related doc cref in `TeamsBaggageBuilder`.
- **Tests** — added unit tests covering the wire → identity mapping and the tenant-only case.

## Verification

- 169 Core + 354 Apps unit tests pass.
- Verified live against a real agentic invoke: the recipient's `tenantId` flowed correctly into the parsed `AgenticIdentity.TenantId`.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/core/src/Microsoft.Teams.Apps/Schema/TeamsChannelAccount.cs
+++ b/core/src/Microsoft.Teams.Apps/Schema/TeamsChannelAccount.cs
@@ -54,7 +54,7 @@ public class TeamsChannelAccount : ChannelAccount
         result.Email = result.Properties.Extract<string>("email");
         result.UserPrincipalName = result.Properties.Extract<string>("userPrincipalName");
         result.UserRole = result.Properties.Extract<string>("userRole");
-        result.TenantId = result.Properties.Extract<string>("tenantId");
+        result.TenantId = channelAccount.TenantId;
         if (string.IsNullOrEmpty(result.AadObjectId))
         {
             result.AadObjectId = result.ObjectId;
--- a/core/src/Microsoft.Teams.Apps/Schema/TeamsChannelAccount.cs
+++ b/core/src/Microsoft.Teams.Apps/Schema/TeamsChannelAccount.cs
@@ -103,4 +103,10 @@ public class TeamsChannelAccount : ChannelAccount
     /// </summary>
     [JsonPropertyName("userRole")]
     public string? UserRole { get; set; }
+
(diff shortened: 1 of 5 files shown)


core/src/Microsoft.Teams.Apps/Schema/TeamsChannelAccount.cs
```diff
@@ -103,10 +103,4 @@ public TeamsChannelAccount()
     /// </summary>
     [JsonPropertyName("userRole")]
     public string? UserRole { get; set; }
-
-    /// <summary>
-    /// Gets or sets the Microsoft Entra tenant ID associated with this account.
-    /// </summary>
-    [JsonPropertyName("tenantId")]
-    public string? TenantId { get; set; }
```
MehakBindra, line 111: result.TenantId = result.Properties.Extract<string>("tenantId"); this line needs to be removed and we need to add result.TenantId  = channelAccount.TenantId ;

MehakBindra (changes requested):

core/src/Microsoft.Teams.Apps/Schema/TeamsChannelAccount.cs
MehakBindra: @copilot fix this

MehakBindra (approved):

corinagum (approved):

PR #593, opened 2026-07-08
feat(apps): streaming error handling and reusable streams for TeamsStreamingWriter

lilyydu:
## Summary

Ports two merged **teams.py** PRs to the .NET `core` streaming writer (`TeamsStreamingWriter`), adapted to its `Append`/`Finalize` API surface:

- microsoft/teams.py#453 — streaming error handling for the 2-minute timeout & "content stream not allowed"
- microsoft/teams.py#495 — support resetting (reusing) response streams

## Streaming error handling (port of #453)

- New exception types in `TeamsStreamingExceptions.cs`: `TerminalStreamException` (base) with `StreamTimedOutException`, `StreamNotAllowedException`, and `StreamCancelledException`.
- The send path now classifies the `403` response body instead of only matching cancellation:
  - `exceeded streaming time` → timed out (flagged, swallowed)
  - `cancel` → cancelled (flagged, swallowed)
  - `not allowed` → `StreamNotAllowedException` (propagates)
  - anything else → `TerminalStreamException` (propagates)
- On a two-minute timeout, `FinalizeResponseAsync` updates the **original streamed message in place** via `UpdateActivityAsync` — it reuses the `streamId` and drops the `streaminfo` entity + stream channel data so the send routes through an update rather than posting a duplicate.
- Cancellation and timeout are swallowed (graceful stop, consistent with the writer's prior cancel behavior); not-allowed and terminal errors surface to the caller.

## Reusable streams (port of #495)

- Appending or sending an informative update after `FinalizeResponseAsync` **reopens the stream on the same writer instance**, starting a new streamed message. The `cancelled` flag stays sticky across reuse.
- `FinalizeResponseAsync` is **idempotent** until the next append/informative update.

## Sample

`core/samples/StreamingBot` gains a `multi-stream` path (no OpenAI required): it streams text, emits an Adaptive Card as the stream 1 final message, finalizes, then reuses the writer for a second streamed message — mirroring the teams.py/teams.ts example.

## Tests

- `TeamsStreamingWriterTests` — 2 existing tests updated to assert reuse/idempotency (previously asserted throw), plus 5 new tests: 403 message → exception mapping, streaming-timeout in-place update, and final-send-timeout fallback to in-place update.
- Full `Microsoft.Teams.Apps.UnitTests` suite passes (359 tests, net10.0); build and `dotnet format` clean on the changed files.
- The `multi-stream` reuse flow was verified end-to-end against a live Teams bot.

## Notes

- Core streaming has no retry wrapper around sends, so #453's `nonRetryable` retry option has no analog here — nothing to port for that piece.
- The Adaptive Card in the sample is built as a raw `JsonObject` because core serializes streamed activities with a source-generated JSON context that does not support the typed `Microsoft.Teams.Cards` models directly.

### DIFF
--- a/core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
+++ b/core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
@@ -310,7 +310,7 @@ public sealed class TeamsStreamingWriter
         {
             _cancelled = true;
             _logger.LogWarning("The streaming was stopped by the user (streamId '{StreamId}').", _streamId);
-            throw new StreamCancelledException(message);
+            throw new StreamCancelledException(message, ex);
         }
     }
 
--- a/core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
+++ b/core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
@@ -36,6 +36,15 @@ namespace Microsoft.Teams.Apps;
 ///     final.AddFeedback(FeedbackTypes.Default);
 ///     await writer.FinalizeResponseAsync(final);
 /// </code>
+///
+/// The writer is reusable: appending or sending an informative update after
+/// <see cref="FinalizeResponseAsync"/> reopens the stream on the same instance and starts a new
+/// streamed message. Finalizing is idempotent until the next append/informative update.
+///
+/// Streaming errors are surfaced per the Teams streaming error codes: cancellation and the
+/// two-minute timeout are handled gracefully (a timed-out stream finalizes by updating the
+/// original message in place), while <see cref="StreamNotAllowedException"/> and other
+/// <see cref="TerminalStreamException"/> errors propagate to the caller.
 /// </remarks>
 public sealed class TeamsStreamingWriter
 {
@@ -51,8 +60,20 @@ public sealed class TeamsStreamingWriter
     private int _sequence;
     private bool _finalized;
     private bool _cancelled;
+    private bool _timedOut;
     private readonly System.Text.StringBuilder _accumulated = new();
     private DateTime _lastChunkSent = DateTime.MinValue;
+
+    /// <summary>
+    /// Whether the stream has been cancelled, for example when the user pressed the Stop button.
+    /// </summary>
+    public bool Cancelled => _cancelled;
+
+    /// <summary>
+    /// Whether streaming exceeded the two-minute limit. When true, the final message is
+    /// sent by updating the original streamed message in place rather than as a streamed chunk.
+    /// </summary>
+    public bool TimedOut => _timedOut;
 
     internal TeamsStreamingWriter(ConversationClient client, TeamsActivity reference, ILogger? logger = null)
     {
@@ -77,12 +98,19 @@ public sealed class TeamsStreamingWriter
     /// </summary>
     public async Task SendInformativeUpdateAsync(string text, CancellationToken cancellationToken = default)
     {
+        // Sending after finalize reopens the stream on the same instance, starting a new streamed message.
+        if (_finalized)
+            ResetForNextStream();
+
+        if (_cancelled)
+            return;
+
         if (_lastChunkSent > DateTime.MinValue)
             throw new InvalidOperationException("Cannot send an informative update after streaming has started.");
 
         _sequence++;
         _logger.LogDebug("Sending informative streaming update (sequence {Sequence}).", _sequence);
-        SendActivityResponse? response = await _client.SendActivityAsync(BuildActivity(text, StreamTypes.Informative), cancellationToken: cancellationToken).ConfigureAwait(false);
+        SendActivityResponse? response = await TrySendChunkAsync(BuildActivity(text, StreamTypes.Informative), cancellationToken).ConfigureAwait(false);
         _streamId ??= response?.Id;
         _logger.LogDebug("Stream started with streamId '{StreamId}'.", _streamId);
     }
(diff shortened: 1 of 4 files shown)


core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -77,12 +98,19 @@ public static TeamsStreamingWriter CreateFromContext<TActivity>(Context<TActivit
     /// </summary>
     public async Task SendInformativeUpdateAsync(string text, CancellationToken cancellationToken = default)
     {
+        // Sending after finalize reopens the stream on the same instance, starting a new streamed message.
+        if (_finalized)
```
heyitsaamir, line 105: log debug?

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -91,38 +119,40 @@ public async Task SendInformativeUpdateAsync(string text, CancellationToken canc
...
     /// </summary>
-    /// <exception cref="InvalidOperationException">Thrown if <see cref="FinalizeResponseAsync"/> has already been called.</exception>
+    /// <remarks>
+    /// Appending after <see cref="FinalizeResponseAsync"/> reopens the stream on the same
+    /// instance, starting a new streamed message.
+    /// </remarks>
     public async Task AppendResponseAsync(string chunk, CancellationToken cancellationToken = default)
     {
+        // Appending after finalize reopens the stream on the same instance, starting a new streamed message.
         if (_finalized)
```
heyitsaamir, line 135: log debug

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -91,38 +119,40 @@ public async Task SendInformativeUpdateAsync(string text, CancellationToken canc
...
-            throw new InvalidOperationException("Cannot append after FinalizeResponseAsync has been called.");
+            ResetForNextStream();
```
heyitsaamir, line 138: we should move this below cancelled.

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -77,12 +98,19 @@ public static TeamsStreamingWriter CreateFromContext<TActivity>(Context<TActivit
     /// </summary>
     public async Task SendInformativeUpdateAsync(string text, CancellationToken cancellationToken = default)
     {
+        // Sending after finalize reopens the stream on the same instance, starting a new streamed message.
+        if (_finalized)
+            ResetForNextStream();
```
heyitsaamir, line 108: move below _cancelled

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -91,38 +119,40 @@ public async Task SendInformativeUpdateAsync(string text, CancellationToken canc
...
-        catch (HttpRequestException ex) when (
-            ex.StatusCode is HttpStatusCode.Gone or HttpStatusCode.NoContent
-            || ex.Message.Contains("Content stream was cancelled", StringComparison.OrdinalIgnoreCase))
-        {
-            _logger.LogWarning("Stream cancelled by user (streamId '{StreamId}').", _streamId);
-            _cancelled = true;
-        }
+        _logger.LogDebug("Sending streaming chunk (sequence {Sequence}, accumulated {Length} chars).", _sequence, _accumulated.Length);
+        SendActivityResponse? response = await TrySendChunkAsync(BuildActivity(_accumulated.ToString(), StreamTypes.Streaming), cancellationToken).ConfigureAwait(false);
+        if (_cancelled || _timedOut)
```
heyitsaamir, line 159: if we time out, we still need to set the stream id right?

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -165,12 +209,158 @@ public async Task FinalizeResponseAsync(MessageActivity? final = null, Cancellat
...
+            catch (StreamTimedOutException)
+            {
+                // The final streamed send tripped the two-minute limit. Update the original
+                // message in place with the buffered content: reuse the id and drop the
+                // stream markers so this routes to an update, not a new streamed chunk.
+                await SendFinalInPlaceAsync(final, cancellationToken).ConfigureAwait(false);
+            }
+            catch (StreamCancelledException)
+            {
+                // Cancelled during the final send; nothing more to send.
```
heyitsaamir, line 239: log?

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -165,12 +209,158 @@ public async Task FinalizeResponseAsync(MessageActivity? final = null, Cancellat
...
+    /// </summary>
+    private async Task<SendActivityResponse?> TrySendChunkAsync(TeamsActivity activity, CancellationToken cancellationToken)
+    {
+        try
+        {
+            return await _client.SendActivityAsync(activity, cancellationToken: cancellationToken).ConfigureAwait(false);
+        }
+        catch (HttpRequestException ex)
+        {
+            try
```
heyitsaamir, line 262: debug log

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
lilyydu: it's already been set previously and when I tested, it was working. but I'll move it up anyway!

corinagum (approved):

heyitsaamir (approved):

PR #594, opened 2026-07-08
Flow bot request context through BotBuilder compat clients

heyitsaamir:
## Summary
- Pass `BotRequestContext` through `CompatUserTokenClient` to all core user token calls.
- Set shared request context for reactive and proactive BotBuilder compat turns.
- Pass request context through remaining compat conversation/update and adapter send paths.
- Add unit coverage for user token, conversation, and adapter request-context plumbing.

## Validation
- `dotnet test core/test/Microsoft.Teams.Apps.BotBuilder.UnitTests/Microsoft.Teams.Apps.BotBuilder.UnitTests.csproj --no-restore`

### DIFF
--- a/core/src/Microsoft.Teams.Apps.BotBuilder/CompatConversations.cs
+++ b/core/src/Microsoft.Teams.Apps.BotBuilder/CompatConversations.cs
@@ -353,7 +353,13 @@ namespace Microsoft.Teams.Apps.BotBuilder
                 coreActivity.ServiceUrl = new Uri(ServiceUrl);
             }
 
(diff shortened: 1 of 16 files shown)


MehakBindra:
Looks good ! Related to the previous PR - https://github.com/microsoft/teams.net/pull/591 can we remove the non BotRequestContext overloads, since it is an optional param anyway

heyitsaamir:
@MehakBindra addressed this in 402f3adb. I removed the redundant non-BotRequestContext overloads from UserTokenClient and made the requestContext parameters optional on the remaining methods. I also updated the internal wrappers/tests to use named cancellationToken arguments where positional calls would otherwise bind to requestContext.

MehakBindra (approved):

PR #596, opened 2026-07-09
Add explicit unauthenticated request bypass

heyitsaamir:
## Summary
- Deprecate `skipAuth` in favor of `DangerouslyAllowUnauthenticatedRequests` while preserving compatibility.
- Allow unauthenticated `/api/messages` requests without parsing a bearer token when the dangerous bypass is enabled in the AspNetCore plugin path. This currently prevents 2.0 from being used for agents playground.
- Support `DangerouslyAllowUnauthenticatedRequests` via `Teams` config/env and the core `TeamsBotApplication` auth path, including `AzureAd` section support.

## Behavior
- When `DangerouslyAllowUnauthenticatedRequests` is enabled, unauthenticated requests are accepted without requiring or parsing the `Authorization` header.
- When `ClientId` is missing and the dangerous bypass is not enabled, requests are rejected with `401` ProblemDetails containing only the generic `Authentication not configured` title/status.
- The remediation guidance to configure `ClientId` or enable `DangerouslyAllowUnauthenticatedRequests` is logged server-side and is not returned in the HTTP response body.

## Validation
- `dotnet test Tests/Microsoft.Teams.Plugins.AspNetCore.Tests/Microsoft.Teams.Plugins.AspNetCore.Tests.csproj --verbosity minimal`
- `dotnet test core/test/Microsoft.Teams.Core.UnitTests/Microsoft.Teams.Core.UnitTests.csproj --verbosity minimal`
- `dotnet build Microsoft.Teams.sln --verbosity minimal`

### DIFF
--- a/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
+++ b/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
@@ -183,7 +183,9 @@ public partial class AspNetCorePlugin : ISenderPlugin, IAspNetCorePlugin
         try
         {
             var request = httpContext.Request;
-            var token = ExtractToken(request);
+            var options = httpContext.RequestServices?.GetService(typeof(AspNetCorePluginOptions)) as AspNetCorePluginOptions;
+            var dangerouslyAllowUnauthenticatedRequests = options?.DangerouslyAllowUnauthenticatedRequests == true;
+            var token = dangerouslyAllowUnauthenticatedRequests ? null : ExtractToken(request);
             var activity = await ParseActivity(request).ConfigureAwait(false);
 
             if (activity is null)
@@ -191,12 +193,14 @@ public partial class AspNetCorePlugin : ISenderPlugin, IAspNetCorePlugin
                 return Results.BadRequest("Missing activity");
             }
 
+            IToken activityToken = token is null ? new UnauthenticatedToken(activity.ServiceUrl) : token;
+
             // Require the token's serviceurl claim to match the activity's serviceUrl
             // (normalized, case-insensitive) when the activity specifies one. Mismatches
             // are logged server-side.
-            if (!string.IsNullOrEmpty(activity.ServiceUrl))
+            if (!dangerouslyAllowUnauthenticatedRequests && !string.IsNullOrEmpty(activity.ServiceUrl))
             {
-                var claimServiceUrl = token.Token.Payload.TryGetValue("serviceurl", out var serviceUrlClaim)
+                var claimServiceUrl = token!.Token.Payload.TryGetValue("serviceurl", out var serviceUrlClaim)
                     ? serviceUrlClaim as string
                     : null;
 
@@ -223,7 +227,7 @@ public partial class AspNetCorePlugin : ISenderPlugin, IAspNetCorePlugin
 
             var res = await Do(new ActivityEvent()
             {
-                Token = token,
+                Token = activityToken,
                 Activity = activity,
                 Extra = data,
                 Services = httpContext.RequestServices
--- a/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
+++ b/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
@@ -6,6 +6,7 @@ using System.Text.Json.Serialization;
 
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
+using Microsoft.IdentityModel.Tokens;
 using Microsoft.Teams.Api.Activities;
 using Microsoft.Teams.Api.Auth;
 using Microsoft.Teams.Api.Clients;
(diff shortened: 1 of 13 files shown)


core/src/Microsoft.Teams.Core/Hosting/BotConfig.cs
```diff
@@ -118,6 +133,25 @@ public static BotConfig Resolve(IServiceCollection services, string sectionName
...
+            return null;
+        }
+
+        if (bool.TryParse(value, out bool result))
+        {
+            return result;
+        }
+
+        throw new InvalidOperationException(
+            $"Configuration value '{section.Path}:{key}' is not a valid boolean: '{value}'.");
```
corinagum, line 159: This behavior diverges from PY and TS, right? Do we want parity?

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/UnauthenticatedToken.cs
```diff
@@ -0,0 +1,22 @@
...
+    public string ServiceUrl { get; } = string.IsNullOrEmpty(serviceUrl)
+        ? "https://smba.trafficmanager.net/teams/"
+        : serviceUrl.EndsWith('/') ? serviceUrl : $"{serviceUrl}/";
+    public CallerType From => CallerType.Azure;
+    public string FromId => "urn:botframework:azure";
+    public DateTime? Expiration => null;
+    public bool IsExpired => false;
+    public IEnumerable<string> Scopes => [];
+    public override string ToString() => string.Empty;
+}
```
corinagum, line 20: This synthesizes trafficmanager when the activity has no service url, but TS and PY do not. Should all three have matching behavior here?

core/src/Microsoft.Teams.Core/Hosting/BotConfig.cs
```diff
@@ -17,6 +17,10 @@ public sealed class BotConfig
 
     internal const string BotFrameworkSectionName = "BotFramework";
 
+    private const string TeamsSectionName = "Teams";
```
singhk97: we don't need a new section imo, we can simply support the key at the root (or default section)

MehakBindra (dismissed):
Can we keep this change to 2.0 ?
EDIT - resolved offline

core/src/Microsoft.Teams.Core/Hosting/BotConfig.cs
MehakBindra: Can we keep only one section ? and the default remains AzureAD

core/src/Microsoft.Teams.Core/Hosting/AuthenticationNotConfiguredHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+            LogLevel.Warning,
+            new EventId(1, "AuthenticationNotConfigured"),
+            "Authentication is not configured. Configure ClientId or enable DangerouslyAllowUnauthenticatedRequests for local development.");
+
+    protected override Task<AuthenticateResult> HandleAuthenticateAsync()
+    {
+        return Task.FromResult(AuthenticateResult.Fail("Authentication not configured"));
+    }
+
+    protected override async Task HandleChallengeAsync(AuthenticationProperties properties)
```
MehakBindra, line 23: This leaves us with no startup-time warning log and defers it to when a request is actually received .

core/src/Microsoft.Teams.Core/Hosting/JwtExtensions.cs
```diff
@@ -79,7 +81,7 @@ public static AuthenticationBuilder AddBotAuthentication(
 
             if (string.IsNullOrWhiteSpace(clientId))
             {
-                builder.AddBypassAuthentication(schemeName, logger);
+                builder.AddAuthenticationNotConfigured(schemeName);
```
singhk97, line 89: same logic should also be applied to this overload

singhk97 (changes requested):

Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.cs
```diff
@@ -183,20 +183,24 @@ public async Task<IResult> Do(HttpContext httpContext, CancellationToken cancell
         try
         {
             var request = httpContext.Request;
-            var token = ExtractToken(request);
+            var options = httpContext.RequestServices?.GetService(typeof(AspNetCorePluginOptions)) as AspNetCorePluginOptions;
+            var dangerouslyAllowUnauthenticatedRequests = options?.DangerouslyAllowUnauthenticatedRequests == true;
+            var token = dangerouslyAllowUnauthenticatedRequests ? null : ExtractToken(request);
             var activity = await ParseActivity(request).ConfigureAwait(false);
```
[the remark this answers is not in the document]
heyitsaamir: now we try catch and throw an unauthorized

core/src/Microsoft.Teams.Core/Hosting/BotConfig.cs
heyitsaamir: aligns now (throws if not a valid boolean we can read)

core/src/Microsoft.Teams.Core/Hosting/BotConfig.cs
heyitsaamir: Removed the section, reused AzureAD

core/src/Microsoft.Teams.Core/Hosting/JwtExtensions.cs
heyitsaamir: fixed

singhk97 (approved):

PR #599, opened 2026-07-10
Add agent lifecycle event handlers

heyitsaamir:
## Summary
- Add typed Agent 365 `agentLifecycle` event activity/value models with `valueType` discrimination for all lifecycle variants.
- Add `EventNames.AgentLifecycle`, lifecycle `valueType`/`eventType` constants, and strongly typed handler registration extensions including `OnAgentLifecycle` and per-variant handlers.
- Preserve generic `OnEvent` compatibility and keep Teams activity metadata when deriving typed event contexts.

## Validation
- `dotnet test core/test/Microsoft.Teams.Apps.UnitTests/Microsoft.Teams.Apps.UnitTests.csproj`
- `dotnet build core/src/Microsoft.Teams.Apps/Microsoft.Teams.Apps.csproj --no-restore`


core/src/Microsoft.Teams.Apps/Handlers/EventHandler.Activity.cs
```diff
@@ -29,12 +32,21 @@ public class EventActivity : TeamsActivity
...
+    public string? ValueType { get; set; }
+
     /// <summary>
     /// Gets or sets the value payload of the event activity.
     /// </summary>
     [JsonPropertyName("value")]
     public JsonNode? Value { get; set; }
 
+    /// <inheritdoc/>
+    public override string ToJson() => CoreActivity.ToJson(this);
```
MehakBindra: why do we need this ? Will we ever send these activities ?

core/src/Microsoft.Teams.Apps/Handlers/AgentLifecycleHandler.cs
```diff
@@ -0,0 +1,262 @@
...
+    /// <param name="app">The Teams bot application.</param>
+    /// <param name="handler">The handler to register.</param>
+    /// <returns>The updated Teams bot application.</returns>
+    public static TeamsBotApplication OnAgentLifecycle(this TeamsBotApplication app, AgentLifecycleHandler handler)
+    {
+        ArgumentNullException.ThrowIfNull(handler, nameof(handler));
+        return app.RegisterAgentLifecycleRoute(
+            EventNames.AgentLifecycle,
+            valueType: null,
+            AgentLifecycleEventActivity.FromEventActivity,
```
MehakBindra: Same, I think this should be AgentLifecycleEventActivity

core/src/Microsoft.Teams.Apps/Handlers/EventHandler.Activity.cs
```diff
@@ -29,12 +32,21 @@ public class EventActivity : TeamsActivity
     [JsonPropertyName("name")]
     public string? Name { get; set; }
 
+    /// <summary>
+    /// Gets or sets the value payload type for event activities that carry typed variants.
+    /// </summary>
+    [JsonPropertyName("valueType")]
+    public string? ValueType { get; set; }
```
MehakBindra: Should this be part of AgentLifecycleEventActivity ?

core/src/Microsoft.Teams.Apps/Handlers/EventHandler.Activity.cs
```diff
@@ -20,7 +20,10 @@ public class EventActivity : TeamsActivity
     public static new EventActivity FromActivity(CoreActivity activity)
     {
         ArgumentNullException.ThrowIfNull(activity);
-        return new EventActivity(activity);
+        EventActivity eventActivity = new(activity);
+        return eventActivity.Name == EventNames.AgentLifecycle
+            ? AgentLifecycleEventActivity.FromEventActivity(eventActivity)
```
MehakBindra: Can we avoid this ? for on event handler we always receive event activity

core/src/Microsoft.Teams.Apps/Handlers/AgentLifecycleHandler.Activity.cs
```diff
@@ -0,0 +1,338 @@
...
+    /// <param name="activity">The source event activity.</param>
+    internal AgentLifecycleEventActivity(EventActivity activity) : base(activity)
+    {
+    }
+}
+
+/// <summary>
+/// Fired when an agentic user identity is created.
+/// </summary>
+public class AgenticUserIdentityCreatedActivity : AgentLifecycleEventActivity<AgenticUserIdentityCreatedValue>
```
MehakBindra: Do we need to define all these classes ? All we need is AgentLifecycleEventActivity<TValue> right ? along with values already defined

MehakBindra (changes requested):

core/src/Microsoft.Teams.Apps/Schema/TeamsActivity.cs
```diff
@@ -68,8 +68,8 @@ internal TeamsActivity(CoreActivity activity) : base(activity)
         base.From = TeamsChannelAccount.FromChannelAccount(activity.From) ?? new TeamsChannelAccount();
         base.Recipient = TeamsChannelAccount.FromChannelAccount(activity.Recipient) ?? new TeamsChannelAccount();
         base.Conversation = TeamsConversation.FromConversation(activity.Conversation) ?? new TeamsConversation();
-        ChannelData = activity.Properties.Extract<TeamsChannelData>("channelData");
```
singhk97, line 69: why are we dropping `activity`

singhk97 (approved):

core/src/Microsoft.Teams.Apps/Handlers/EventHandler.Activity.cs
MehakBindra: Any thoughts on this ?

heyitsaamir: thought about it some more - you're right. let me move it

core/src/Microsoft.Teams.Apps/Schema/TeamsActivity.cs
heyitsaamir: Chatted with @MehakBindra   looks like there's a bug that this uncovered. Keeping the activity actually modifies the incoming activity instead of the activity copied by passing to base.

MehakBindra (approved):

PR #615, opened 2026-07-23
Refine Agent 365 SDK identity naming and scoping

heyitsaamir:
## Summary

This keeps `AgenticUser` as the concrete Teams/activity-facing identity model while making Core `AgenticIdentity` the SDK operation scope for request, proactive send, API-client, and auth plumbing. The carrier is created from `AgenticUser` today and gives the SDK one canonical seam for future Agent 365 scopes without spreading new overloads through every helper.

## Decisions

- **Concrete model vs operation scope.** `AgenticUser` remains the activity domain object because Teams payloads still describe an agentic user. `AgenticIdentity` is a disciplined discriminated carrier for SDK operations, so low-level plumbing depends on the scope abstraction rather than one concrete subject type.
- **Wire contracts stay service-owned.** Activity JSON keys, lifecycle aliases/valueTypes, role literals, telemetry tags, and dependency-boundary names remain unchanged, including `agenticUserId`, `agenticAppId`, `agenticAppBlueprintId`, `agenticAppInstanceId`, lifecycle `AgenticUser...` / `agenticUser...`, role `agenticUser`, auth flow tag `agentic`, service-owned `agentIdentityBlueprintId`, and upstream `WithAgentUserIdentity(...)`.
- **Future app-instance path.** `AgenticIdentityKind.AgenticAppInstance` is present only as the future carrier shape. Adding real app-instance behavior should be one factory/wrapper at the canonical seam, not a broad overload set across sends and clients.

## Deferred

Actual `AgenticAppInstance` behavior/model support is future work; this PR only preserves the operation-scope shape needed to add it cleanly later.

## Reviewer tips

Start at Core schema and request plumbing (`ChannelAccount`, `AgenticUser`, `AgenticIdentity`, `BotRequestContext`, `BotAuthenticationHandler`), then Apps API-client scoping/proactive send helpers, lifecycle handlers, and the Agent365/ObservabilityBot/PABot samples and focused tests.

## Testing

Focused Core activity/request tests passed, including `AgenticIdentityTests`, `BotRequestContextTests`, `UserTokenClientTests`, `CoreActivityTests`, and `ConversationClientTests` on net8.0/net10.0. Focused Apps lifecycle/activity/diagnostics tests and BotBuilder `TeamsApiClientTests` passed; changed Agent365, ObservabilityBot, PABot, and CompatBot samples built successfully; `IntegrationTests` built successfully; `git diff --check` passed. Earlier live Teams smoke test passed with `core/samples/Agent365/Agent365.csproj` and Teams agentic user `Demo Test 2`, confirming inbound parsing and outbound echo behavior without exposing secrets.


singhk97 (approved):
lgtm

corinagum (approved):

core/src/Directory.Build.props
```diff
@@ -20,6 +20,7 @@
     <PropertyGroup>
         <AnalysisLevel>latest-all</AnalysisLevel>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1903</WarningsNotAsErrors>
```
MehakBindra: Revert this ?

MehakBindra (changes requested):

core/src/Microsoft.Teams.Apps/Clients/MemberClient.cs
```diff
@@ -20,24 +20,24 @@ public class MemberClient
...
     private readonly Uri _serviceUrl;
-    private readonly AgenticIdentity? _agenticIdentity;
+    private readonly AgenticUser? _agenticUser;
 
-    internal MemberClient(Uri serviceUrl, CoreConversationClient client, AgenticIdentity? agenticIdentity = null)
+    internal MemberClient(Uri serviceUrl, CoreConversationClient client, AgenticUser? agenticUser = null)
     {
         _serviceUrl = serviceUrl;
         _client = client;
-        _agenticIdentity = agenticIdentity;
```
MehakBindra, line 29: Agentic id contains app id and user id right ? (and agentic id feels analogous to bot id) , I'm not sure if we should change this

heyitsaamir: So according Anand, 
Agentic Identity is a confusing term, and we should not use it.
AgenticUser means it has an app instance id (aka app id) AND a user id.

Does that make sense? Let me know if you want to chat about this.

core/src/Microsoft.Teams.Core/Http/BotRequestContext.cs
```diff
@@ -53,15 +53,27 @@ public record BotRequestContext
...
+
+        return new BotRequestContext { AgenticIdentity = agenticIdentity };
+    }
+
+    /// <summary>
+    /// Builds context carrying only the supplied agentic user.
+    /// </summary>
+    /// <param name="agenticUser">The agentic user.</param>
+    /// <returns>The context carrying an <see cref="AgenticIdentity"/> created from <paramref name="agenticUser"/>.</returns>
+    public static BotRequestContext FromAgenticUser(AgenticUser agenticUser)
```
MehakBindra, line 75: Let's not have so many methods, I think we should create agentic user or agentic app instance from channel account and agentic id from those. So many convenience  wrappers also become confusing, wdyt ?

core/src/Microsoft.Teams.Core/Schema/AgenticIdentity.cs
```diff
@@ -4,46 +4,94 @@
...
+    /// Gets the kind of agentic identity.
     /// </summary>
-    public string? AgenticUserId { get; set; }
+    public AgenticIdentityKind Kind { get; }
 
     /// <summary>
-    /// Tenant ID associated with the agentic identity.
+    /// Gets the agentic app instance ID.
     /// </summary>
-    public string? TenantId { get; set; }
```
heyitsaamir, line 23: this needs to come back

core/src/Microsoft.Teams.Core/Schema/AgenticUser.cs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Teams.Core.Schema;
+
+/// <summary>
+/// Represents an agentic user for user-delegated token acquisition.
+/// </summary>
+public sealed class AgenticUser
```
MehakBindra, line 9: Aamir do we need this class ? Isn't AgenticIdentity with "Kind" enough ?

MehakBindra (changes requested):

heyitsaamir:
Closing per the updated Agent 365 identity direction: C# stays as-is, and the AgenticIdentity-only follow-up work is continuing in Python/TypeScript.

PR #629, opened 2026-07-29
Thread streamed activities under the inbound message via replyToId

lilyydu:
## What

Ports [microsoft/teams.ts#646](https://github.com/microsoft/teams.ts/pull/646) to teams.net.

`TeamsStreamingWriter` now sets `ReplyToId` to the inbound activity id on **every** outbound send

Covered send paths:
- Informative updates (`SendInformativeUpdateAsync`)
- Intermediate streaming chunks (`AppendResponseAsync`)
- The final message (`FinalizeResponseAsync`), including the in-place `UpdateActivity` fallback taken after the two-minute streaming timeout

The TS PR applies `replyToId: this.ref.activityId` centrally in `send()`. The .NET streamer has no single send choke point (routing is passed per call), so `ReplyToId = _reference.Id` is set in the two spots that construct outbound activities: `BuildActivity` (informative + streaming) and `FinalizeResponseAsync` (final + timeout in-place update).

## Testing

- **Unit tests**: added two tests to `TeamsStreamingWriterTests` — one asserting informative/streaming/final all carry `replyToId` = inbound id, and one asserting the post-timeout in-place `PUT` is also threaded. Full `TeamsStreamingWriterTests` suite passes (net10.0).
- **Live E2E**: ran the `StreamingBot` sample against real Teams behind a devtunnel with trace logging. Confirmed **14/14** outbound activity sends (informative, every streaming chunk, and final) carried `replyToId` equal to the corresponding inbound activity id.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
+++ b/core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
@@ -193,6 +193,7 @@ public sealed class TeamsStreamingWriter
 
         final ??= new MessageActivityInput();
         final.Text ??= _accumulated.ToString();
+        final.ReplyToId = _reference.Id;
 
         if (string.IsNullOrEmpty(final.Text) && (final.Attachments == null || final.Attachments.Count == 0))
             throw new InvalidOperationException(
@@ -374,9 +375,12 @@ public sealed class TeamsStreamingWriter
 
     private StreamingActivityInput BuildActivity(string text, StreamType streamType)
     {
-        return StreamingActivityInput.CreateBuilder()
+        StreamingActivityInput activity = StreamingActivityInput.CreateBuilder()
             .WithText(text)
             .WithStreamInfo(streamType, _streamId, _sequence)
             .Build();
+
+        activity.ReplyToId = _reference.Id;
+        return activity;
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
```diff
@@ -374,9 +375,12 @@ private void ResetForNextStream()
 
     private StreamingActivityInput BuildActivity(string text, StreamType streamType)
     {
-        return StreamingActivityInput.CreateBuilder()
+        StreamingActivityInput activity = StreamingActivityInput.CreateBuilder()
             .WithText(text)
             .WithStreamInfo(streamType, _streamId, _sequence)
             .Build();
+
+        activity.ReplyToId = _reference.Id;
```
MehakBindra, line 383: Lily can you add it to the Streaming activity builder first and then use that ?

MehakBindra (changes requested):

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
lilyydu: @MehakBindra That would only cover the informative and streaming chunks, but not the final activity. I don't think we should have it in the builder also b/c `replyToId` is from the inbound context, and the builder is generic reference-agnostic

MehakBindra (approved):

core/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs
MehakBindra: We'll review later for streamingto use fluent methods/builder properly

PR #633, opened 2026-08-03
fix: omit sequence from final stream messages

corinagum:
## Summary

- clear stale channel `StreamSequence` when final stream metadata is added
- omit the stream entity sequence property when no sequence is supplied
- add serialized payload regression coverage

Follow-up to microsoft/teams.ts#688.

## Testing

- `dotnet format core/core.slnx --include core/src/Microsoft.Teams.Apps/Schema/Entities/StreamInfoEntity.Extensions.cs core/test/Microsoft.Teams.Apps.UnitTests/MessageActivityInputTests.cs --verify-no-changes --no-restore`
- `dotnet test core/test/Microsoft.Teams.Apps.UnitTests/Microsoft.Teams.Apps.UnitTests.csproj -f net10.0 -v minimal --no-restore` (484 passed)

### DIFF
--- a/core/src/Microsoft.Teams.Apps/Schema/Entities/StreamInfoEntity.Extensions.cs
+++ b/core/src/Microsoft.Teams.Apps/Schema/Entities/StreamInfoEntity.Extensions.cs
@@ -40,18 +40,18 @@ public static class StreamInfoEntityExtensions
 
         activity.ChannelData.StreamId = resolvedStreamId;
         activity.ChannelData.StreamType = streamType;
-        if (streamSequence.HasValue)
-        {
-            activity.ChannelData.StreamSequence = streamSequence.Value;
-        }
+        activity.ChannelData.StreamSequence = streamSequence;
 
         activity.Entities ??= [];
         StreamInfoEntity entity = new()
         {
             StreamId = resolvedStreamId,
-            StreamType = streamType,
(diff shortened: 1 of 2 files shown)


core/src/Microsoft.Teams.Apps/Schema/Entities/StreamInfoEntity.Extensions.cs
```diff
@@ -40,18 +40,18 @@ internal static StreamInfoEntity AddToActivity(TeamsActivityInput activity, Stre
 
         activity.ChannelData.StreamId = resolvedStreamId;
         activity.ChannelData.StreamType = streamType;
-        if (streamSequence.HasValue)
-        {
-            activity.ChannelData.StreamSequence = streamSequence.Value;
-        }
+        activity.ChannelData.StreamSequence = streamSequence;
```
MehakBindra, line 43: Is this after some guidance to not set on Channel Data ?

corinagum: https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux?pivots=teams-sdk-csharp#stream-message-through-rest-api

More details in original teams.ts PR

heyitsaamir (approved):

PR #645, opened 2026-08-06
Feature: Receive files in personal scope

corinagum:
Adds API for receiving files to the app in personal scope. A user uploads a file, which arrives on the activity, and `context.Files.ListAsync()` provides metadata of the file(s) and lazy handlers to download it. The url is `tempauth` and will expire.

- **`context.Files`**: `ListAsync()` / `FirstAsync()` map inbound `file.download.info` attachments to `IncomingFile`. Never throws: non-message activities and empty/malformed attachments yield an empty list; bad entries are skipped.
- **`IncomingFile`**: file metadata with lazy `DownloadAsync()`, `StreamAsync()`, `TextAsync(encoding?)`, `SaveAsAsync(path)`. Each read re-fetches (bounded by the short-lived URL).
- **`DownloadedFile`**: buffered snapshot you own (`Bytes`, `ContentType`, `Filename`, `SourceUrl`), with `Text()` / `SaveAsAsync()`.
- **Errors**: Differentiates between first fetch error (url expired before first fetch), reread error (url expired after first fetch), and scope not supported error (today, only 1:1).
- **Types**: new `FileDownloadInfo` schema + `file.download.info` content-type constant (a new `AttachmentContentType` member); reuses the existing `ConversationType` (already has `Channel`) for scope. `FileDownloader` is public so it can be resolved from DI; `TeamsBotApplication`'s constructor takes an optional `Files.FileDownloader`, registered as a typed `HttpClient` by `AddTeamsBotApplication`.

## Testing

- Unit tests: `FilesAccessorTests.cs`, `IncomingFileTests.cs`, `TeamsBotApplicationHostingExtensionsTests.cs`
- Manual e2e

## Note

- After investigating thumbnail url behavior, I adjusted `TeamsAttachment.ThumbnailUrl` docstrings to indicate it will not be populated on incoming files.
- Similarly for `FileDownloadInfo.Etag`, this field may be set by Graph, which is not in scope for this PR.

### DIFF
--- a/src/Microsoft.Teams.Apps/Files/FileDownload.cs
+++ b/src/Microsoft.Teams.Apps/Files/FileDownload.cs
@@ -1,0 +1,152 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Teams.Apps.Schema;
+
+namespace Microsoft.Teams.Apps.Files;
+
+/// <summary>
+/// A freshly opened, single-consumption byte stream plus the metadata resolved while opening it. Owns the underlying <see cref="HttpResponseMessage"/>; disposing this stream releases the response and connection. Read-only and non-seekable: it hands back the raw response body without leaking the response.
+/// </summary>
+internal sealed class OpenedFileStream : Stream
+{
+    private readonly Stream _inner;
+    private readonly HttpResponseMessage? _response;
+
+    public OpenedFileStream(Stream inner, Uri sourceUrl, string contentType, HttpResponseMessage? response = null)
+    {
+        _inner = inner;
+        SourceUrl = sourceUrl;
+        ContentType = contentType;
+        _response = response;
+    }
+
+    /// <summary>The URL the bytes were actually fetched from.</summary>
+    public Uri SourceUrl { get; }
+
+    /// <summary>MIME type resolved from the response, falling back to the incoming file's.</summary>
+    public string ContentType { get; }
+
+    public override bool CanRead => _inner.CanRead;
+    public override bool CanSeek => false;
+    public override bool CanWrite => false;
+    public override long Length => _inner.Length;
+
+    public override long Position
+    {
+        get => _inner.Position;
+        set => throw new NotSupportedException();
+    }
+
+    public override int Read(byte[] buffer, int offset, int count) => _inner.Read(buffer, offset, count);
+
+    public override int Read(Span<byte> buffer) => _inner.Read(buffer);
+
+    public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        => _inner.ReadAsync(buffer, offset, count, cancellationToken);
+
+    public override ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default)
+        => _inner.ReadAsync(buffer, cancellationToken);
+
+    public override void Flush()
+    {
+    }
+
+    public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();
+
+    public override void SetLength(long value) => throw new NotSupportedException();
+
+    public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            _inner.Dispose();
+            _response?.Dispose();
+        }
+
+        base.Dispose(disposing);
+    }
+
+    public override async ValueTask DisposeAsync()
+    {
+        await _inner.DisposeAsync().ConfigureAwait(false);
+        _response?.Dispose();
+        await base.DisposeAsync().ConfigureAwait(false);
+    }
+}
+
+/// <summary>
+/// Opens byte streams for inbound files, keyed on conversation scope so every scope's receive path extends this one place rather than branching in callers.
+/// </summary>
+internal static class FileDownload
+{
+    private static readonly HttpClient DefaultHttpClient = new();
+
+    /// <summary>Open a byte stream for an inbound file. Only <c>personal</c> is implemented; other scopes throw <see cref="FileScopeNotSupportedException"/> until their Graph receive path lands.</summary>
+    public static Task<OpenedFileStream> OpenFileStreamAsync(
+        ConversationType? scope,
+        Uri? downloadUrl,
+        string? contentType,
+        bool priorFetchSucceeded,
+        HttpClient? httpClient,
+        CancellationToken cancellationToken)
+    {
+        if (scope == ConversationType.Personal)
+        {
+            return OpenPersonalFileStreamAsync(downloadUrl, contentType, priorFetchSucceeded, httpClient, cancellationToken);
+        }
+
+        throw new FileScopeNotSupportedException(scope?.Value ?? "unknown");
+    }
+
+    private static async Task<OpenedFileStream> OpenPersonalFileStreamAsync(
+        Uri? downloadUrl,
+        string? contentType,
+        bool priorFetchSucceeded,
+        HttpClient? httpClient,
+        CancellationToken cancellationToken)
+    {
+        if (downloadUrl is null)
+        {
+            throw new InvalidOperationException("cannot download personal file: no download URL is available");
+        }
+
+        HttpClient client = httpClient ?? DefaultHttpClient;
+
+        // Plain GET with no bearer token: the download URL embeds its own `tempauth` credential, and attaching a
+        // credential can get the request rejected.
+        HttpResponseMessage response = await client
+            .GetAsync(downloadUrl, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
+            .ConfigureAwait(false);
+
+        try
+        {
+            int status = (int)response.StatusCode;
+
+            if (status is 401 or 403)
+            {
+                throw new FileUrlExpiredException(priorFetchSucceeded ? FileUrlExpiredReason.Reread : FileUrlExpiredReason.FirstFetch);
+            }
+
+            if (!response.IsSuccessStatusCode)
+            {
+                throw new HttpRequestException($"failed to download file: {status} {response.ReasonPhrase}".Trim());
+            }
+
+            string resolvedContentType = response.Content.Headers.ContentType?.ToString()
+                ?? contentType
+                ?? "application/octet-stream";
+
+            Stream stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
+
+            return new OpenedFileStream(stream, downloadUrl, resolvedContentType, response);
+        }
+        catch
+        {
+            response.Dispose();
+            throw;
+        }
+    }
+}
--- a/src/Microsoft.Teams.Apps/Files/FileDownload.cs
+++ b/src/Microsoft.Teams.Apps/Files/FileDownload.cs
@@ -113,6 +113,11 @@ internal static class FileDownload
             throw new InvalidOperationException("cannot download personal file: no download URL is available");
         }
 
+        if (!downloadUrl.Scheme.Equals(Uri.UriSchemeHttps, StringComparison.OrdinalIgnoreCase))
+        {
+            throw new InvalidOperationException("cannot download file: download URL must use https");
+        }
+
         HttpClient client = httpClient ?? DefaultHttpClient;
 
         // Plain GET with no bearer token: the download URL embeds its own `tempauth` credential, and attaching a
--- a/src/Microsoft.Teams.Apps/Files/FileDownload.cs
+++ b/src/Microsoft.Teams.Apps/Files/FileDownload.cs
@@ -98,7 +98,7 @@ internal static class FileDownload
             return OpenPersonalFileStreamAsync(downloadUrl, contentType, priorFetchSucceeded, httpClient, cancellationToken);
         }
 
-        throw new FileScopeNotSupportedException(scope?.Value ?? "unknown");
+        throw new FileScopeNotSupportedException(scope);
     }
 
     private static async Task<OpenedFileStream> OpenPersonalFileStreamAsync(
--- a/src/Microsoft.Teams.Apps/Files/FileDownload.cs
+++ b/src/Microsoft.Teams.Apps/Files/FileDownload.cs
@@ -30,11 +30,11 @@ internal sealed class OpenedFileStream : Stream
     public override bool CanRead => _inner.CanRead;
     public override bool CanSeek => false;
     public override bool CanWrite => false;
-    public override long Length => _inner.Length;
+    public override long Length => throw new NotSupportedException();
 
     public override long Position
     {
-        get => _inner.Position;
+        get => throw new NotSupportedException();
         set => throw new NotSupportedException();
     }
 
--- a/src/Microsoft.Teams.Apps/Files/FilesAccessor.cs
+++ b/src/Microsoft.Teams.Apps/Files/FilesAccessor.cs
@@ -1,0 +1,128 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using Microsoft.Extensions.Logging;
+using Microsoft.Teams.Apps.Schema;
+
+namespace Microsoft.Teams.Apps.Files;
+
+/// <summary>
+/// Accessor for the uploaded files on the current inbound activity, exposed as <c>ctx.Files</c>. Reads the files attached to the current inbound activity and exposes them as lazy <see cref="IncomingFile"/> handles.
+/// <para>"Files" is the uploaded-file view over the raw <c>ctx.Activity.Attachments</c> list. Uploaded files arrive as attachments where <c>ContentType</c> is <c>file.download.info</c>, carrying file metadata (a <c>downloadUrl</c> plus identifiers) rather than the bytes themselves, which are fetched from that URL. This accessor maps each to an <see cref="IncomingFile"/>, and skips everything else in <c>Attachments</c> (adaptive cards, mentions, other non-file content) as well as malformed file entries, never throwing. For each file it returns, the original wire attachment (the metadata object, not the bytes) is retained on <see cref="IncomingFile.Raw"/>. A malformed or non-file attachment is reachable only through the raw <c>Activity.Attachments</c> list.</para>
+/// <para>This covers the file-upload path, not "any uploaded media". What matters is how the content arrived, not the file's MIME type, so file <em>type</em> is unrestricted (pdf, docx, png, etc.) as long as it was sent as an uploaded file. An image sent as a file appears here, but the same image pasted inline does not.</para>
+/// </summary>
+public sealed class FilesAccessor
+{
+    private readonly TeamsActivity _activity;
+    private readonly ILogger _logger;
+
+    /// <summary>Initializes a new instance of the <see cref="FilesAccessor"/> class for the given inbound activity.</summary>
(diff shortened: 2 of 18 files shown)


src/Microsoft.Teams.Apps/Files/FileDownload.cs
```diff
@@ -0,0 +1,152 @@
...
+                throw new HttpRequestException($"failed to download file: {status} {response.ReasonPhrase}".Trim());
+            }
+
+            string resolvedContentType = response.Content.Headers.ContentType?.ToString()
+                ?? contentType
+                ?? "application/octet-stream";
+
+            Stream stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
+
+            return new OpenedFileStream(stream, downloadUrl, resolvedContentType, response);
```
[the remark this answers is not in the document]
corinagum: value is stable

src/Microsoft.Teams.Apps/Files/FileDownload.cs
```diff
@@ -0,0 +1,157 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Teams.Apps.Schema;
+
+namespace Microsoft.Teams.Apps.Files;
+
+/// <summary>
+/// A freshly opened, single-consumption byte stream plus the metadata resolved while opening it. Owns the underlying <see cref="HttpResponseMessage"/>; disposing this stream releases the response and connection. Read-only and non-seekable: it hands back the raw response body without leaking the response.
+/// </summary>
+internal sealed class OpenedFileStream : Stream
```
MehakBindra: Do we need to subclass `Stream` here ?  should we return a simple `OpenedFile` result object containing the `Stream`, `ContentType`, and `SourceUrl`?

like

internal sealed class OpenedFile : IAsyncDisposable
{
    public required Stream Stream { get; init; }
    public required string ContentType { get; init; }
    public required Uri SourceUrl { get; init; }
    internal required HttpResponseMessage Response { get; init; }

    public async ValueTask DisposeAsync()
    {
        await Stream.DisposeAsync();
        Response.Dispose();
    }
}

With composition, callers explicitly use `openedFile.Stream`, so no reimplementation/ forwarding to "inner" code is needed. `OpenedFile` clearly owns that stream and the HTTP response and disposes both.

src/Microsoft.Teams.Apps/Files/FilesAccessor.cs
```diff
@@ -0,0 +1,128 @@
...
+            Source = FileSource.BotActivity,
+            // Maps the wire's `contentUrl` (a browsable link to the file in OneDrive/SharePoint) to `WebUrl`; not fetchable like `downloadUrl`.
+            WebUrl = attachment.ContentUrl,
+            Raw = attachment,
+            DownloadUrl = downloadUrl,
+        };
+    }
+
+    /// <summary>Coerce an attachment's <c>Content</c> (already-typed, a <see cref="JsonElement"/> from the wire, or another object) into a <see cref="FileDownloadInfo"/>. Returns <c>null</c> on malformed content so the caller can skip it rather than throw.</summary>
+    private static FileDownloadInfo? AsFileDownloadInfo(object? content)
```
MehakBindra, line 117: instead of doing this, since this is part of the payload, can we introduce ContentType-based serialization/deserialization in TeamsAttachment itself ?

src/Microsoft.Teams.Apps/Files/FilesAccessor.cs
```diff
@@ -0,0 +1,128 @@
...
+        IList<IncomingFile> files = await ListAsync(cancellationToken).ConfigureAwait(false);
+        return files.Count > 0 ? files[0] : null;
+    }
+
+    /// <summary>Derive the conversation scope from the inbound activity.</summary>
+    private ConversationType DetectScope()
+        => _activity.Conversation?.ConversationType ?? ConversationType.Personal;
+
+    /// <summary>Map a single activity attachment to an <see cref="IncomingFile"/>, or <c>null</c> when the attachment is not an uploaded file or is malformed. Never throws: unusable attachments are skipped so one bad entry cannot drop the rest.</summary>
+    private IncomingFile? ToIncomingFile(TeamsAttachment attachment, int index, ConversationType scope)
```
MehakBindra: same, should be owned by incoming file , and can be renamed TryParseIncomingFile

src/Microsoft.Teams.Apps/Files/FilesAccessor.cs
MehakBindra: Similar to Entities

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
```diff
@@ -0,0 +1,108 @@
...
+    public Uri? WebUrl { get; init; }
+
+    /// <summary>The raw underlying attachment/graph object for escape-hatch access.</summary>
+    public object? Raw { get; init; }
+
+    /// <summary>Short-lived, pre-authorized download URL (personal scope).</summary>
+    internal Uri? DownloadUrl { get; init; }
+
+    /// <summary>Injectable HTTP client, defaulting to a shared client; used to keep tests off the network.</summary>
+    internal HttpClient? HttpClient { get; init; }
```
MehakBindra: `HttpClient` should belong to the downloader, not `IncomingFile`

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Text;
+using Microsoft.Teams.Apps.Schema;
+
+namespace Microsoft.Teams.Apps.Files;
+
+/// <summary>
+/// A lazy handle to a file attached to the current inbound activity.
+/// <para>Nothing is downloaded until a byte method is called. The handle stays live and holds no memoized bytes, so each of <see cref="StreamAsync"/>/<see cref="DownloadAsync"/>/<see cref="TextAsync"/>/<see cref="SaveAsAsync"/> fetches afresh. For a personal file that re-fetch is bounded by the short-lived download URL lifetime and may hit its expiry; to read the same file several ways, call <see cref="DownloadAsync"/> once and reuse the returned <see cref="DownloadedFile"/>.</para>
+/// </summary>
+public sealed class IncomingFile
```
MehakBindra, line 13: `IncomingFile` has required invariants—name, scope, source, download URL, and downloader—so a constructor is better than scattered `required`/`internal init` properties.

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// <summary>Display name including extension when known.</summary>
+    public required string Name { get; init; }
+
+    /// <summary>
+    /// The file's MIME type, when Teams provides it with the file. Files received today do not include one, so
+    /// this is usually <c>null</c>; the type resolved from the download response is on the returned
+    /// <see cref="DownloadedFile"/>, not written back here.
+    /// </summary>
+    public string? ContentType { get; init; }
```
MehakBindra, line 43: should we remove this then ? if it always null here

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
MehakBindra, line 13: Once content is typed as FileDownloadInfo, wondering if we even need this data model ?
`TeamsAttachment` already contains the inbound metadata; once `Content` is typed as `FileDownloadInfo`, `IncomingFile` duplicates much of it.

MehakBindra (dismissed):

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
MehakBindra: Nevermind, realized we probably need it for other paths like direct graph ? I'm working on the app graph part of .NET , wondering if we should do that first

src/Microsoft.Teams.Apps/Files/FileDownload.cs
```diff
@@ -0,0 +1,157 @@
...
+        await base.DisposeAsync().ConfigureAwait(false);
+    }
+}
+
+/// <summary>
+/// Opens byte streams for inbound files, keyed on conversation scope so every scope's receive path extends this one place rather than branching in callers.
+/// </summary>
+internal static class FileDownload
+{
+    private static readonly HttpClient DefaultHttpClient = new();
```
MehakBindra: need to provide a non null Http client via constructor and use DI, this seems to be an anti pattern

src/Microsoft.Teams.Apps/Files/FileDownload.cs
```diff
@@ -0,0 +1,157 @@
...
+
+    /// <summary>Open a byte stream for an inbound file. Only <c>personal</c> is implemented; other scopes throw <see cref="FileScopeNotSupportedException"/> until their Graph receive path lands.</summary>
+    public static Task<OpenedFileStream> OpenFileStreamAsync(
+        ConversationType? scope,
+        Uri? downloadUrl,
+        string? contentType,
+        bool priorFetchSucceeded,
+        HttpClient? httpClient,
```
MehakBindra: HttpClient? httpClient, should be removed

src/Microsoft.Teams.Apps/Files/FileDownload.cs
```diff
@@ -0,0 +1,157 @@
...
+        await _inner.DisposeAsync().ConfigureAwait(false);
+        _response?.Dispose();
+        await base.DisposeAsync().ConfigureAwait(false);
+    }
+}
+
+/// <summary>
+/// Opens byte streams for inbound files, keyed on conversation scope so every scope's receive path extends this one place rather than branching in callers.
+/// </summary>
+internal static class FileDownload
```
MehakBindra: rename to FileDownloader ?

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
corinagum: No, it's possible to be hydrated from Graph calls, which is out of scope for this PR.

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
corinagum: I'm keeping just `DownloadUrl` as an initializer. It isn't a universal invariant, it's specific to how files are fetched on the activity body. Graph will be off `UniqueId`, so requiring `DownloardUrl` would be meaningless in other scopes.

src/Microsoft.Teams.Apps/Files/FilesAccessor.cs
corinagum: Seven of the eight `AttachmentContentType` values are cards, and `Microsoft.Teams.Apps` doesn't reference `Microsoft.Teams.Cards`. Typing content by ContentType needs one of:

- `Apps` takes a dependency on `Cards`. Legal, since `Cards` has no project references of its own so there's no cycle, but it means every Apps consumer pulls it.
- A registry where Cards self-registers its content types, keeping the dependency pointed the right way.

The narrow version, typing just `file.download.info` content, is small enough to do here. The general version depends on the Apps/Cards layering call, which is bigger than this PR and needs to be decided on before being actionable.

src/Microsoft.Teams.Apps/Files/FileDownload.cs
corinagum: Mostly done. `OpenedFile` record now carries `Stream`, `SourceUrl`, `ContentType`, and metadata is off the stream type. The subclass doesn't disappear, though. `StreamAsync` is public and returns `Task<Stream>`. `ResponseOwningStream` releases `HttpResponseMessage` when the caller disposes it. Pulled from Azure Storage implementation as precedence. 

`CA2000` now fires twice, which I've suppressed with justifications.

src/Microsoft.Teams.Apps/TeamsBotApplication.cs
```diff
@@ -114,7 +125,8 @@ public TeamsBotApplication(
         IHttpContextAccessor httpContextAccessor,
         ILogger<TeamsBotApplication> logger,
         TeamsBotApplicationOptions? options = null,
-        TurnStateLoader? stateLoader = null)
+        TurnStateLoader? stateLoader = null,
+        IHttpClientFactory? httpClientFactory = null)
```
MehakBindra: Need to review

MehakBindra: It should take FileDownloader in the ctor OR we should add it directly in AddTeamsBotApplication .

src/Microsoft.Teams.Apps/TeamsBotApplication.cs
```diff
@@ -88,6 +88,16 @@ public OAuthFlow GetOAuthFlow(string connectionName)
...
 
+    /// <summary>
+    /// Opens byte streams for inbound files, backing <c>ctx.Files</c>. Its client comes from
+    /// <see cref="IHttpClientFactory"/> when the app is built through the hosting extensions.
+    /// </summary>
+    internal Files.FileDownloader FileDownloader { get; }
+
+    // Fallback for an app constructed directly rather than through DI, where no factory is available. Shared and
+    // never disposed, which is the supported lifetime for a long-lived HttpClient.
+    private static readonly HttpClient FallbackFileDownloadClient = new();
```
MehakBindra: remove

MehakBindra: direct, non-DI construction is not a supported scenario

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
MehakBindra: I think we need to review this carefully as apps and graph being tightly coupled in .NET is not a given

src/Microsoft.Teams.Apps/TeamsBotApplication.cs
corinagum: Taking `FileDownloader` in the constructor doesn't compile:
```
error CS0051: Inconsistent accessibility: parameter type 'FileDownloader' is less
accessible than method 'TeamsBotApplication.TeamsBotApplication(...)'
```
`FileDownloader` is internal and the constructor is public. To do it we'd have to make `FileDownloader` public, and `OpenedFile` with it since it's the return type, which puts two types users never touch into the public surface.
AddTeamsBotApplication still can't reach the public constructor for the same accessibility reason, so it doesn't remove the `IHttpClientFactory` parameter by itself.

Two ways forward:
1. Make `FileDownloader` (and `OpenedFile`) public, register the typed client, inject it directly. Cleanest DI, wider public API.
1. Keep both internal and keep the `IHttpClientFactory` parameter.

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
corinagum: Nothing in this PR takes a dependency on Graph. `UniqueId` is carried because the payload provides it, and the only current consumer is the personal download path, which uses `DownloadUrl`.

I'd rather not design the Graph seam in this PR.

corinagum: Outside personal scope there is no download URL, so Graph is required by the platform regardless of packaging; `UniqueId` is the locator that path keys off.

src/Microsoft.Teams.Apps/TeamsBotApplication.cs
corinagum: Tried it. Dropping the field and using the factory directly fails the build on our own analyzer:
`error CA1062: In externally visible method 'TeamsBotApplication.TeamsBotApplication(...)',
validate parameter 'httpClientFactory' is non-null before using it.`

So it has to become required. Adding `ArgumentNullException.ThrowIfNull(httpClientFactory)` builds, and then 141 of 511 unit tests fail. Every one of them constructs the app directly:  TeamsBotApplicationTests.cs:69 ,  ConversationUpdateHandlerTests.cs:200 ,  CreateDerivedContextTests.cs:169 ,  HandlerStatePropagationTests.cs:561 ,  AgentLifecycleHandlerTests.cs:393 ,  ContextStateTests.cs:77 .

I agree direct construction isn't a supported product scenario. But it's how the suite is written today, so removing the fallback means either passing a factory through six test helpers or accepting a required parameter on a public constructor.

What I'd suggest instead: replace the static with a per-instance new HttpClient() when no factory is supplied. That drops the global, keeps the constructor's public shape, and keeps the suite green. One client per app rather than one per process is fine given apps are singletons.

If you'd rather make it required and fix up the tests, I'll do that, but it's a breaking change to a public, shipped constructor and I'd want it to be deliberate.

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
MehakBindra: I'd rather not design the Graph seam in this PR. - agree, but a lot of this is designed considering graph will also be using this API right ?We can merge this but before moving forward I would like to do a POC with graph and maybe you could help me understand a little bit how graph will be used ?

src/Microsoft.Teams.Apps/Files/FileDownload.cs
MehakBindra: Okay I see, then I think what you had before was better?

src/Microsoft.Teams.Apps/TeamsBotApplication.cs
MehakBindra: I don't think we should have the factory in the ctor

src/Microsoft.Teams.Apps/Files/FilesAccessor.cs
MehakBindra: Fair

src/Microsoft.Teams.Apps/Files/IncomingFile.cs
corinagum: There's no Graph package in today, so a Graph POC means building that integration, not validating against an existing one. This design proposal and organization of PRs is very intentionally keeping Graph scope out of this. The design review handled the overal public API (list and download) and was approved. This is a large scope increase gating a feature that is complete and tested for the scope it claims. Python already merged this shape (#550) and TypeScript is open on it (#712), so holding .NET forks the API across the three SDKs.
The Graph seam is already reserved: `OpenFileStreamAsync` switches on scope and throws `FileScopeNotSupportedException` for everything but personal. A Graph path is an added branch, not a redesign of what's here.

src/Microsoft.Teams.Apps/TeamsBotApplication.cs
```diff
@@ -109,12 +117,17 @@ public OAuthFlow GetOAuthFlow(string connectionName)
     /// }
     /// </code>
     /// </example>
+    [SuppressMessage(
```
MehakBindra: We should not have this on our main app class. we need to think this through

corinagum: The original code used a static readonly HttpClient field, which CA2000 exempts. Removing that per your earlier comment forced a per-instance new HttpClient(), which CA2000 does flag, hence the suppression. Moved the default client into `FileDownloader` behind `CreateDefault()`.

src/Microsoft.Teams.Apps/Files/FileDownload.cs
corinagum: Reverted..

singhk97 (approved):

singhk97 (commented):
can we showcase usage in a sample as well

PR #651, opened 2026-08-13
Add AIFileAnalysisBot sample

corinagum:
Adds `AIFileAnalysisBot`, a sample for the file-receive API from #645. It reads files attached in a personal (1:1) chat and sends the ones it understands to Azure OpenAI.

One `OnMessage` handler covers two paths:

- **Basic (no LLM)** replies with an Adaptive Card for files the sample cannot analyze, showing the metadata the file API exposes and the bytes that were downloaded.
- **AI** converts text files and images into model content parts, sends them in a single request, and streams the analysis back.

The flow is `context.Files.ListAsync()`, then `DownloadAsync()` once per file, then `FileContext.Classify` to sort each download into `Text`, `Image`, or `Unsupported`. `Classify` falls back to the file's extension when the MIME type does not match, because Teams often misreports source files (`.ts` comes through as `video/vnd.dlna.mpeg-tts`). Image bytes are inlined as a data URI rather than passing the model the download URL, which carries a short-lived `tempauth` credential.

Where a block is labeled, the label is either `FILE RECEIVE` (the SDK API, the part worth copying) or `SAMPLE GUARDRAIL` (limits this sample picked). The guardrails: 5 files per message, 100 KB text per file, 250 KB text per message, 1 MB per image, PNG/JPEG/GIF/WebP only, and no conversation state. Anything dropped or truncated sends the user a message saying why.

## Prerequisites

The manifest needs `supportsFiles: true` on the bot entry. Without it Teams does not enable the attachment UI in the bot's chat, so there is nothing to receive.

## Testing

`dotnet build` succeeds with 0 warnings and 0 errors, and `dotnet format --verify-no-changes` is clean.

Manually tested against a real bot in the Teams web client:

| Case | Result |
| --- | --- |
| No attachment | Prompts for a file |
| Text `.txt` | Analyzed and streamed |
| JSON `.json` | Analyzed and streamed |
| Image `.png` | Analyzed and streamed |
| Two images in one message | Both files reached the model in one request |
| Mixed text + image in one message | Both files reached the model in one request |
| Unsupported `.zip` | Basic card, no model call |
| 192 KB text file | Truncated at the 100 KB cap with a warning |

Other languages:

- TypeScript: microsoft/teams.ts#727
- Python: microsoft/teams.py#562

### DIFF
--- a/samples/AIFileAnalysisBot/Program.cs
+++ b/samples/AIFileAnalysisBot/Program.cs
@@ -16,20 +16,40 @@ using Microsoft.Teams.Apps.Files;
 WebApplicationBuilder builder = WebApplication.CreateSlimBuilder(args);
 builder.Services.AddTeamsBotApplication();
 
-string endpoint = builder.Configuration["AzureOpenAI:Endpoint"] ?? throw new InvalidOperationException("AzureOpenAI:Endpoint is required.");
-string apiKey = builder.Configuration["AzureOpenAI:ApiKey"] ?? throw new InvalidOperationException("AzureOpenAI:ApiKey is required.");
-string deployment = builder.Configuration["AzureOpenAI:Deployment"] ?? throw new InvalidOperationException("AzureOpenAI:Deployment is required.");
+// SAMPLE GUARDRAIL: the file API needs no model, so the sample stays usable without Azure OpenAI settings. Without
+// them it answers every file with the metadata card instead of analyzing it, which keeps download, content type,
+// scope, and source demonstrable with no model subscription.
+string? endpoint = builder.Configuration["AzureOpenAI:Endpoint"];
+string? apiKey = builder.Configuration["AzureOpenAI:ApiKey"];
+string? deployment = builder.Configuration["AzureOpenAI:Deployment"];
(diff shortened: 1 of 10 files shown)


samples/AIFileAnalysisBot/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+using Microsoft.Teams.Apps.Files;
+
+// Two kinds of code live in this sample, labeled throughout:
+//
+// - FILE RECEIVE is the Teams SDK file API itself. This is the part worth copying into your own app.
+// - SAMPLE GUARDRAIL is this sample deciding what it is willing to forward to a model. Those limits are arbitrary
+//   product choices, not SDK requirements, and your app should pick its own.
+WebApplicationBuilder builder = WebApplication.CreateSlimBuilder(args);
+builder.Services.AddTeamsBotApplication();
+
```
singhk97, line 18: AI file analysis is a great showcase, but requiring OpenAI/Azure OpenAI keys to run the sample can be restrictive. When users don't provide keys, we can fall back to a "basic" flow that returns Adaptive Card metadata instead. This lets people without model API keys still try out the Files API features.

singhk97 (approved):
lgtm!

PR #655, opened 2026-08-17
 Don't require selectedChannel when deserializing channel data

corinagum:
`TeamsChannelDataSettings.SelectedChannel` was declared `required`, and `System.Text.Json` enforces `required` members (.NET 7+). A `channelData.settings` of `{}` therefore threw:
```
JsonException: JSON deserialization for type 'Settings' was missing required properties including: 'selectedChannel'.
```
Found while investigating teams.py#563, where the Teams service sent `channelData.app  as  {}` . .NET was otherwise unaffected; every other channel-data property is already nullable, but `SelectedChannel` was the one strict field and would fail the same way.

This is latent rather than actively reported: `settings` appears less often than `app`. Fixing it now avoids the same class of outage.

## Change
 `public required TeamsChannel SelectedChannel`  >  `public TeamsChannel? SelectedChannel`

Inbound models shouldn't enforce required fields on service-provided data. No consumers of `SelectedChannel` exist in  `src/`, `test/`, or `Samples/` .

## Tests
New `TeamsChannelDataDeserializationTests.cs` empty `app` / `channel` / `team` / `tenant` / `settings`  all deserialize,  `SelectedChannel`  is null for empty settings, and populated values round-trip.

## Validation
- Build: 0 warnings, 0 errors
- 522 tests pass (net10.0)
- No net new format issues

### DIFF
--- a/src/Microsoft.Teams.Apps/Schema/TeamsChannelData.cs
+++ b/src/Microsoft.Teams.Apps/Schema/TeamsChannelData.cs
@@ -37,7 +37,7 @@ public class TeamsChannelDataSettings
     /// <summary>
     /// Selected channel information.
     /// </summary>
-    [JsonPropertyName("selectedChannel")] public required TeamsChannel SelectedChannel { get; set; }
+    [JsonPropertyName("selectedChannel")] public TeamsChannel? SelectedChannel { get; set; }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


singhk97 (approved):
nice proactive fix. can you elaborate on the scenario where this field is omitted

PR #660, opened 2026-08-26
Fix targeted activity integration tests

corinagum:
## What changed

**Fix `CreateMessageActivity(text, recipient)` silently dropping its recipient.** The helper in `ApiClientTests.cs` accepted a `ChannelAccount` and then returned the activity without ever applying it, so targeted activities went out with no recipient. The service rejects those with `400 BadArgument: Recipient of Targeted Message cannot be null`, which failed `Activities_CreateTargetedAsync`, `Activities_UpdateTargetedAsync`, and `Activities_DeleteTargetedAsync`. The helper now calls `WithRecipient(..., isTargeted: true)`.

**Add a regression test for targeted update payloads.** `UpdateActivityAsync_WithIsTargeted_DoesNotSendRecipient` asserts that the serialized body omits `recipient`. C# already behaves correctly here, but only by caller convention: `ConversationClient` serializes the activity verbatim and does no stripping, so nothing was guarding the behavior. This closes a parity gap with the equivalent TypeScript coverage.

## Testing

Integration suite run against a live canary tenant: the three targeted activity tests now pass, with no change to any other test outcome. The new unit test passes and was mutation-checked by injecting a recipient, which fails it as expected.

No product code changed.

### DIFF
--- a/test/IntegrationTests/README.md
+++ b/test/IntegrationTests/README.md
@@ -9,8 +9,9 @@ This project runs integration tests against Teams Server using bot and agentic i
   - Bot app registration (client ID + secret)
   - Agentic app registration (client ID + secret) — optional
   - A team with at least one channel
-  - A scheduled meeting
-  - At least 2 test users in the conversation
+  - At least 3 non-bot users in the test conversation (`GroupChat_ThreeMembers` needs three)
+  - A regular scheduled meeting (not a channel meeting) with the bot app installed *in the meeting*
+  - An OAuth connection on the Azure Bot resource named to match `TEST_CONNECTION_NAME`
 
 ## RunSettings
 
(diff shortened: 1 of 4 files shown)


test/IntegrationTests/README.md
```diff
@@ -90,12 +102,27 @@ dotnet test IntegrationTests/IntegrationTests.csproj \
...
-| `Bots` | 2 | Sign-in URL and resource |
-| `Users` | 3 | Token get, status, sign-out |
+| `Users` | 5 | Sign-in URL and resource, token get, status, sign-out |
 | `Client` | 1 | ForServiceUrl scoped client |
 | `Diagnostic` | 13 | Conversation creation matrix |
 | `ErrorHandling` | 3 | Error cases (compat layer) |
 
+> These account for 68 of the suite's 72 tests; the remaining 4 carry no `Category` trait. A filter naming a category that does not exist prints `No test matches the given testcase filter` and **exits 0**, so a typo looks exactly like a clean pass.
+
+## Troubleshooting
```
MehakBindra, line 112: nice!

test/IntegrationTests/README.md
```diff
@@ -9,8 +9,9 @@ This project runs integration tests against Teams Server using bot and agentic i
   - Bot app registration (client ID + secret)
   - Agentic app registration (client ID + secret) — optional
   - A team with at least one channel
-  - A scheduled meeting
-  - At least 2 test users in the conversation
+  - At least 3 non-bot users in the test conversation (`GroupChat_ThreeMembers` needs three)
```
MehakBindra, line 12: curious, why we need 3?

MehakBindra (approved):

test/IntegrationTests/README.md
corinagum: 2 regular users + a bot = 1:1 human chat, not group chat.

PR #669, opened 2026-09-10
feat(apps): resolve Agentic User file bytes through Graph /shares

corinagum:
### What this does

Agentic User file attachments arrive with a `contentUrl` and **no** `downloadUrl`, so the pre-authorized ("tempauth") URL traditional bots rely on is absent. This resolves those bytes through Microsoft Graph, `GET /shares/{u!encoded}/driveItem/content`, using the Agentic User's own credential.

The route is chosen from the **shape of the attachment**, not from the identity: a `contentUrl` with no `downloadUrl`, at personal scope, resolves via `graphShare`. Everything else keeps the existing `preauthUrl` path unchanged.

### What it deliberately does not do

**An expired pre-authorized URL is terminal.** It raises `FileUrlExpiredError` and makes no Graph call.

A developer can still call Graph with their own client, and keeps the locator because `contentUrl` is public API. The constraint is on what the SDK does automatically.

Auditing trap: the agentic token is itself delegated-shaped, carrying `scp` rather than `roles`. Follow the identity, not the word "delegated".

### Why `.default` and not a named scope

A by-name request hard-fails on exact-string consent mismatch: a blueprint consented to `Files.ReadWrite.All` returns `AADSTS65001` and **no token at all** for a `Files.Read.All` request. Graph's own `/shares` docs list `Files.ReadWrite.All` as least-privileged and list no read-only permission, so admins following Microsoft's guidance land exactly where a by-name request breaks.

The cost is a late, ambiguous 403 when a file permission is missing. We pay that in the SDK instead: the pre-flight rejects a token carrying permissions but **none file-capable**, which restores an early, legible failure. A `Files.Read.All`-then-`.default` ladder was rejected, because `AADSTS65001` is also what "no file permission consented at all" produces, so the ladder makes the most common deployment error harder to diagnose, not easier.

**For security review:** under `.default` the token carries the blueprint's full consented set, including unrelated scopes. The mitigation is the grant, not the request string: do not consent unrelated scopes on a file-handling blueprint.

### Testing

Unit and integration coverage is in the diff. All three scenarios were also run **live against real Teams**, tenant asserted twice per run (browser session state, and the SharePoint host in the bot's own log).

| Scenario | Result |
| --- | --- |
| Traditional bot, fresh file | `preauthUrl`, byte-exact |
| Agentic User via Graph | `/shares/{u!}/driveItem/content` 200, byte-exact, driving the shipped helpers |
| Expired URL | `FileUrlExpiredError`, **zero requests to `graph.microsoft.com` in the entire process log** |

The third row was checked by grepping the whole bot process log for an *absence*, not by reading the rendered result, and the expiry is real rather than simulated: the same `UniqueId` returns 200 early and a hard 401 later. And the pre-authorized URL lifetime is **not a constant**: three runs bounded the flip at (62.7, 72.4], (62.5, 68.4] and (63.6, 66.4] minutes. Poll rather than picking a wait time.

---

## This language

22 commits, 39 files, +3592 / -91.

**A compatibility overload is kept deliberately.** C# bakes optional arguments into the **caller**, so removing the five-parameter `TeamsBotApplication` signature is a runtime `MissingMethodException` for anything compiled against the old shape, even though the source change is compatible. Verified against the real `v2.1.0` tag: that constructor exists there, and it is the only at-risk shipped signature in the change. It is hidden from IntelliSense and pinned by a reflection assertion against emitted metadata, because a positional call would simply rebind to the six-parameter overload and prove nothing.

`ExpiredPreauthUrl_IsTerminal_AndDoesNotReachForGraph` drives the shipped downloader with an expired `downloadUrl` **and** a valid `contentUrl`, the locator a fallback would use, and asserts zero token acquisitions. Supplying the `contentUrl` is the point: the URL is terminal because there is no recovery, not because the test withheld the means to recover.

**`Microsoft.Teams.Graph` does not exist and is not this PR's job.** .NET uses the official Graph package.

**Running the integration tests.** `test/IntegrationTests` is deliberately not referenced by `Microsoft.Teams.slnx`, so solution-wide runs skip it. Invoke it by project path.


src/Microsoft.Teams.Apps/Files/FileDownloader.cs
```diff
@@ -109,17 +114,22 @@ public override async ValueTask DisposeAsync()
 /// Initializes a new instance of the <see cref="FileDownloader"/> class.
 /// </remarks>
 /// <param name="httpClient">Client used to fetch file bytes. Supplied by DI; must not be null.</param>
-public sealed class FileDownloader(HttpClient httpClient)
+/// <param name="logger">Logger used to record which route produced the bytes. Which route ran is otherwise invisible from outside, and the two routes fail in different ways, so a developer diagnosing a download has no way to tell them apart without it.</param>
+public sealed class FileDownloader(HttpClient httpClient, ILogger<FileDownloader>? logger = null)
 {
```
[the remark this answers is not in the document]
corinagum: The type has not been shipped yet.

src/Microsoft.Teams.Apps/Files/FileDownloader.cs
```diff
@@ -109,17 +114,22 @@ public override async ValueTask DisposeAsync()
 /// Initializes a new instance of the <see cref="FileDownloader"/> class.
 /// </remarks>
 /// <param name="httpClient">Client used to fetch file bytes. Supplied by DI; must not be null.</param>
-public sealed class FileDownloader(HttpClient httpClient)
+/// <param name="logger">Logger used to record which route produced the bytes. Which route ran is otherwise invisible from outside, and the two routes fail in different ways, so a developer diagnosing a download has no way to tell them apart without it.</param>
+public sealed class FileDownloader(HttpClient httpClient, ILogger<FileDownloader>? logger = null)
```
lilyydu, line 118: this breaks prev compiled customers I believe - we should also have a one-parameter forwarding constructor

src/Microsoft.Teams.Core/Hosting/AddBotApplicationExtensions.cs
```diff
@@ -163,6 +164,13 @@ public static IServiceCollection EnsureMsalServices(this IServiceCollection serv
         ArgumentNullException.ThrowIfNull(botConfig);
         ArgumentNullException.ThrowIfNull(botConfig.MsalConfigurationSection);
 
+        // The same acquisition the outbound pipeline performs, reachable by callers that are not an HTTP pipeline.
+        // TryAdd because this method is documented as safe to call more than once.
+        services.TryAddSingleton(sp => new BotTokenProvider(
+            sp.GetRequiredService<IAuthorizationHeaderProvider>(),
+            botConfig.SectionName,
```
lilyydu, line 173: this auto defaults to the first bot/auth registered which may end up being the wrong app later on - maybe we should register `BotTokenProvider` as a keyed singleton where the key is the auth section name?

src/Microsoft.Teams.Apps/Files/FileDownloader.cs
corinagum: `FileDownloader` never shipped, so it's non-breaking.

MehakBindra (approved):

lilyydu (approved):

PR #682, opened 2026-09-23
feat(apps): add Socket Mode protocol transport

teddyam:
## Summary

This is the bottom layer of the Socket Mode work. It adds the internal protocol transport foundation only:

- Socket Mode protocol models and JSON serialization
- envelope parsing and reply construction
- negotiation client and coverage
- internal connection contract

This layer is internal-only. It does not add a public `App` option and does not change application behavior.

## Validation

Focused `SocketModeProtocolTests`, `SocketModeEnvelopeTests`, and `SocketModeNegotiatorTests`:

- `net8.0`: 50 passed
- `net10.0`: 50 passed


test/Microsoft.Teams.Apps.UnitTests/SocketMode/SocketModeEnvelopeTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        Assert.True(found);
+        Assert.Equal("payload-activity", activity?.Id);
+    }
+
+    [Theory]
+    [InlineData("\"not-an-object\"")]
+    [InlineData("[]")]
+    [InlineData("""{ "id": "missing-type" }""")]
+    [InlineData("""{ "type": 42 }""")]
+    public void TryReadActivity_FallsBackWhenPayloadIsMalformed(string malformedPayload)
```
lilyydu: I think these four cases fail at the 'TryReadActivityCandidate' step - could we add one more `[InlineData("""{ "type": "message", "id": 42 }""")]` so it makes it into the try/catch

src/Microsoft.Teams.Apps/SocketMode/SocketModeProtocol.cs
```diff
@@ -21,5 +21,5 @@ internal static class SocketModeProtocol
     /// <summary>
     /// The path for the Socket Mode negotiate endpoint.
     /// </summary>
-    internal const string NegotiatePath = "/v3/websockets/connect";
+    internal const string NegotiatePath = "/amer/v3/websockets/connect";
```
lilyydu: oo wait this can be apac/emea also

teddyam: oh bruh i swear i changed that yeah

lilyydu (approved):

PR #683, opened 2026-09-23
feat(apps): add SignalR Socket Mode connection

teddyam:
## Summary

Stacked on #682.

- add the official SignalR client adapter and a one-generation Socket Mode connection
- gate activity processing on `SocketReady` and return activity client results
- handle terminal closes, cancellation, explicit stop, and disposal
- add focused Socket Mode tests across both configured target frameworks

This layer intentionally does not add public `App` integration.


test/Microsoft.Teams.Apps.UnitTests/SocketMode/SignalRSocketConnectionTests.cs
```diff
@@ -0,0 +1,433 @@
...
+
+        Assert.Equal(1, harness.Negotiator.CallCount);
+        Assert.Equal(NegotiateUri, harness.Negotiator.NegotiateUri);
+        Assert.Equal(SignalRUri, harness.SignalRBuilder.Url);
+        Assert.Equal("signalr-token", harness.SignalRBuilder.AccessToken);
+        Assert.Equal(TimeSpan.FromSeconds(15), harness.SignalRBuilder.KeepAliveInterval);
+        Assert.Equal(TimeSpan.FromSeconds(30), harness.SignalRBuilder.ServerTimeout);
+        Assert.Equal(TimeSpan.FromHours(1), harness.Connection.TokenLifetime);
+        Assert.Equal(1, harness.SignalR.StartCount);
+        Assert.Equal(1, harness.ReadyFrames.Count);
```
lilyydu: nit: xUnit2013 build warning here — can do `Assert.Single(harness.ReadyFrames)`

src/Microsoft.Teams.Apps/SocketMode/SignalRClientConnection.cs
```diff
@@ -0,0 +1,130 @@
...
+                "SocketReady",
+                handler));
+    }
+
+    public void OnClosed(Action<Exception?> handler)
+    {
+        ArgumentNullException.ThrowIfNull(handler);
+        ThrowIfDisposed();
+
+        _connection.Closed += exception =>
```
lilyydu: should we consider keeping the delegate here, and removing it in DisposeAsync? that wait itll unsubscribe and follows `OnActivity` and `OnReady`

src/Microsoft.Teams.Apps/SocketMode/SignalRSocketConnection.cs
```diff
@@ -0,0 +1,286 @@
...
+
+    [SuppressMessage(
+        "Design",
+        "CA1031:Do not catch general exception types",
+        Justification = "Cleanup failure is logged so the original startup exception remains the reported failure.")]
+    private async Task StopAfterFailedStartAsync()
+    {
+        try
+        {
+            await StopAsync(CancellationToken.None).ConfigureAwait(false);
```
lilyydu: should we consider disposing the HubConnection here to match TS/PY?

src/Microsoft.Teams.Apps/SocketMode/SignalRSocketConnection.cs
```diff
@@ -0,0 +1,286 @@
...
+        }
+        catch (Exception exception)
+        {
+            _logger.LogWarning(
+                exception,
+                "Socket Mode ready observer failed after the connection became ready.");
+        }
+    }
+
+    private void HandleClosed(Exception? error)
```
lilyydu: for PY, I passed an explicit planned flag here cause a stop vs a real drop arrive as `error: null` (may try to reconnect as a result)

teddyam: good catch! added `planned` flag to distinguish the two cases

src/Microsoft.Teams.Apps/SocketMode/SignalRSocketConnection.cs
teddyam: yes yes fixed

src/Microsoft.Teams.Apps/SocketMode/SignalRClientConnection.cs
teddyam: closed handler is treated a lil differently bc its a conn lifecycle event not invoked by the server. i do agree it should follow the same delegation and removal as the other handlers tho so i made the change to DisposeAsync

lilyydu (approved):

